Issue #3, opened 2017-01-04
Specifying a conda environment


dhimmel:
It would be great to create a conda environment for this project. The environment should use Python 2.7 rather than 3.6 (see https://github.com/HazyResearch/snorkel/issues/509).

In private communication, @ajratner wrote:

> In particular, if you're willing to absorb the bumps of code being actively developed, you can check out the dev branch: https://github.com/HazyResearch/snorkel/tree/dev

Also see @ajratner's [note here](https://thinklab.com/discussion/brainstorming-future-directions-for-hetionet/227#note-284):

> However, it's certainly still at a stage where our active involvement–both for help/bugfixes and active collaboration on code development–seems to be a big help, in part since the code is evolving so rapidly in response to feedback. We're hoping that it will begin to increasingly stabilize soon; and any feedback you have for us on how we could make it easier to use would be greatly appreciated, either via github issues, a skype call, or on this platform!

@danich1, for these reasons I think we may want to use the a dev version of snorkel. You can specify this in the conda `environment.yml` using a `pip` installation. Example at http://stackoverflow.com/a/32799944/4651668.

dhimmel:
You'll want to include the packages in https://github.com/HazyResearch/snorkel/blob/dev/python-package-requirement.txt, which currently contains:

```
bs4
futures
ipywidgets
jupyter
lxml
matplotlib
nltk
numba
numbskull
numpy>=1.11
pandas
requests
scipy>=0.18
sklearn
sqlalchemy>=1.0.14
theano>=0.8.2
```

Most of these are in the anaconda channel. But the ones that are not, can be installed with `pip`.

danich1:
Got it. So far have all installed just working on that bug I mentioned earlier.

ajratner:
As a heads up we're probably going to merge a big PR into dev, and then soon after merge this into master as a new release of Snorkel, within the next 1-2 weeks; just so you know!  Any other issues here with setup?

danich1:
No issues to report now. Excited for the update!

dhimmel:
Closed by https://github.com/greenelab/snorkeling/pull/4 in 549ef4509ce6b886ae9628fe94acd595c9559aa1.

PR #4, opened 2017-01-10
Conda environment 

danich1:
This pull request is for fixing the issue #3. I updated the environment.yml and provided installation instructions in the README.md. @dhimmel let me know what you think. I don't foresee any issues with the environment.yml, but fingers crossed everything goes well.


environment.yml
```diff
@@ -0,0 +1,27 @@
+name: snorkeling
+channels: !!python/tuple
+- defaults
+dependencies:
+- _license=1.1=py27_1
+- enum34=1.1.6=py27_0
+- llvmlite=0.13.0=py27_0
+- numba=0.28.1=np111py27_0
```
dhimmel: Let's remove the build information (e.g. `=np111py27_0`) as they often create OS-specific incompatibility.

environment.yml
```diff
@@ -0,0 +1,27 @@
...
+- pip:
```
dhimmel, line 23: Most of these packages you're installing with pip should be available from the default anaconda channel.

You can try `conda search package`, so `conda search futures` and it will tell you what versions are available for what versions of python.

environment.yml
```diff
@@ -0,0 +1,27 @@
...
+    - bs4
```
dhimmel: Generally package names are same in anaconda channel as PyPI (where `pip` gets its packages), but here you need to use `beautifulsoup4` for the package in dependencies.

environment.yml
```diff
@@ -0,0 +1,27 @@
...
+    - futures
+    - ipywidgets
+    - jupyter
+    - lxml
+    - matplotlib
+    - nltk
+    - git+https://github.com/HazyResearch/numbskull@master
```
dhimmel: Let's specify a specific version of `numbskull` like a commit or release tag.

environment.yml
```diff
@@ -0,0 +1,27 @@
...
+    - numpy>=1.7
+    - pandas
+    - requests
+    - scipy>=0.18
+    - sklearn
+    - sqlalchemy>=1.0.14
+    - theano>=0.8.2
+    - enum34==1.1.6
+    - numba
+prefix: /home/davidnicholson/anaconda3/envs/snorkeling
```
dhimmel: Remove as install specific.

environment.yml
dhimmel: Use similar method to specify `snorkel` -- currently snorkel appears not to be in this specification.

set_env.sh
```diff
@@ -0,0 +1,7 @@
+export SNORKELHOME="$HOME/Documents/snorkel"
```
dhimmel: Hardcoded path not ideal. Why are setting these environment variables necessary?

environment.yml
```diff
@@ -0,0 +1,27 @@
+name: snorkeling
+channels: !!python/tuple
```
dhimmel: `channels` can be removed for now since we're just using the default.

README.md
```diff
@@ -2,6 +2,38 @@
...
+```
+
+##Running The Project
+After modifying the environment script please type in the following command: 
+
+```bash 
+bash run.sh
+```  
+
+_Note_: Before running the script above please edit the following script [set_env.sh](set_env.sh). Change the `SNORKELHOME` variable to point to the snorkel directory you cloned above. Depending on where this repository is located you may have to change the `WORKINGPATH` variable. If you have to change it, make sure it points to this repository.
```
dhimmel: Is there a reason why snorkel must be installed using this method rather than just using pip? Ideally, we could free ourselves from having to deal with these paths.

README.md
```diff
@@ -2,6 +2,38 @@
 
 A pilot project to extract medical indications (when a compound treats a disease) from the literature using [Snorkel](https://github.com/HazyResearch/snorkel).
 
+##Installing/Setting Up The Conda Environment
```
dhimmel: Great work with the README and documentation.

Minor markdown comments (stylistic not functional).

+ space after heading pounds: `## Installing/Setting Up The Conda Environment`
+ blank line after heading

dhimmel (changes requested):
Awesome first Greene Lab pull request. Some :taco:s coming your way.

The goal here is that we create a reproducible computational environment. Therefore I think we want to specify versions of each package. This way, if you reinstall the environment on another machine, you won't get different versions of a package.

Some more comments in the line by line.

set_env.sh
danich1: Short answer: They are necessary to run snorkel.

Long answer: The way snorkel is set up is that they are setting the snorkel path as a global variable. They do this in order to set up the python path for module inclusion/checking to see if stanford's corenlp is installed. I don't know a way around it.

README.md
danich1: So going back to my earlier comment. Snorkel hasn't reached the point where it has become a stand alone package (no setup.py). Fingers crossed they create a new one. (*addition to Snorkel's todo list right..?) Right now I just cloned their repo to avoid unnecessary duplication of files.

danich1:
I made the necessary changes. Let me know if I missed anything.

environment.yml
```diff
@@ -0,0 +1,43 @@
...
+- requests=2.11.1 
+- scipy=0.18.1 
+- scikit-image=0.12.3 
+- scikit-learn=0.17.1 
+- sqlalchemy=1.0.13 
+- enum34=1.1.6 
+- numba=0.28.1 
+- beautifulsoup4=4.5.1
+- pip:
+    - bs4
```
dhimmel: Delete packages here that are installed above.

environment.yml
```diff
@@ -0,0 +1,43 @@
+name: snorkeling
+dependencies:
```
dhimmel, line 2: Could we sort these alphabetically?

run.sh
```diff
@@ -0,0 +1,21 @@
+# Set & move to home directory
+source set_env.sh
+cd "$SNORKELHOME"
+
+# Make sure the submodules are installed
+git submodule update --init --recursive
```
dhimmel, line 4: What submodule is this for? Want to make sure this won't update the submodule beyond the version specified by the repo.

set_env.sh
dhimmel: Okay let's see if we can remove the user-specific parts. If snorkel is a submodule, you can get its absolute path with

```sh
`pwd`/snorkel
```

README.md
dhimmel: Got it. I think the right move is to include snorkel as a submodule. See [an example](https://github.com/dhimmel/drugcentral) of what a submodule looks like on GitHub.

run.sh
danich1: Upon closer inspection this submodule is for the treedlib repo. I don't think there should be a version issue because on the master branch they specify a commit hash in their submodule.

danich1:
Round two for changes. One question I have is for the submodules should I specify branch or are we going to stick with commit hashes?

run.sh
```diff
@@ -0,0 +1,25 @@
+# Set & move to home directory
+source set_env.sh
+#Make sure the snorkel submodule is installed
+git submodule update --init --recursive
+
+#Change directory into snorkel
+cd "$SNORKELHOME"
+
+# Make sure the submodules are installed
+git submodule update --init --recursive
```
dhimmel: Is this second `git submodule update` as the `--recursive` above should take care of subsubmodules?

environment.yml
```diff
@@ -0,0 +1,26 @@
...
+- numpy=1.11.1 
+- pandas=0.18.1 
+- requests=2.11.1 
+- scikit-image=0.12.3
+- scikit-learn=0.17.1
+- scipy=0.18.1  
+- sqlalchemy=1.0.13 
+- pip:
+    - git+https://github.com/HazyResearch/numbskull@40ac1af20538c17b4726c963c69adcb81314efa5
+    - theano>=0.8.2
```
dhimmel: Lock down version: `theano==0.8.2`

environment.yml
```diff
@@ -0,0 +1,26 @@
+name: snorkeling
+dependencies:
+- _license=1.1
```
dhimmel: Curious, where does the `_license` dependency come from?

set_env.sh
```diff
@@ -0,0 +1,7 @@
+export SNORKELHOME="`pwd`/snorkel"
+echo "Snorkel home directory: $SNORKELHOME"
+export PYTHONPATH="$PYTHONPATH:$SNORKELHOME:$SNORKELHOME/treedlib"
+export PATH="$PATH:$SNORKELHOME:$SNORKELHOME/treedlib"
+export WORKINGPATH="$HOME/Documents/snorkeling"
```
dhimmel: Should this be

```sh
export WORKINGPATH=`pwd`
```

dhimmel (commented):
I was able to clone the repo and install the conda environment on Mac -- fantastic. Also was able to run `run.sh` which operated as intended, without error.

Right now submodule contains the latest master commit HazyResearch/snorkel@e49ecb2047327173ed56b2f85d4a71bdc4487bbc. Not sure if we wanted to use the latest dev commit as discussed in https://github.com/greenelab/snorkeling/issues/3. Not a huge issue as we can always update later when needed.

environment.yml
danich1: Its not needed from what I gathered. I ran `conda env export` and that was included in the dependencies.

README.md
```diff
@@ -2,6 +2,40 @@
...
+
+## Running The Project
+
+After modifying the environment script please type in the following command: 
+
+```bash 
+bash run.sh
+```  
+
+_Note_: Before running the script above please edit the following script [set_env.sh](set_env.sh). Change the `SNORKELHOME` variable to point to the snorkel directory you cloned above. Depending on where this repository is located you may have to change the `WORKINGPATH` variable. If you have to change it, make sure it points to this repository.
```
dhimmel: Most of this note is no longer relevant, I believe.

dhimmel (approved):
Pro PR!

Issue #5, opened 2017-01-23
Update snorkel to v0.5


dhimmel:
The big update @ajratner mentioned is now merged into master.

@danich1, let's update the snorkel submodule to HazyResearch/snorkel@36e8bcb69227e3903a8574eee775af784a53e635 (approx v0.5).

dhimmel:
@danich1, protip -- in your commit message, you can write any of the following:

```
Closes https://github.com/greenelab/snorkeling/issues/5
Closes greenelab/snorkeling#5
Closes #5
```

Which will automatically close this issue once the commit is merged into master.

PR #6, opened 2017-01-24
Upgrade environment to use snorkel v0.5

danich1:
Closes https://github.com/greenelab/snorkeling/issues/5

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -22,4 +22,4 @@ dependencies:
 - sqlalchemy=1.1.4
 - pip:
     - git+https://github.com/HazyResearch/numbskull@40ac1af20538c17b4726c963c69adcb81314efa5
-    - theano==0.8.2
+    - tensorflow==0.12.1
--- a/environment.yml
+++ b/environment.yml
@@ -9,7 +9,7 @@ dependencies:
 - jupyter_console=5.0.0 
 - jupyter_core=4.2.0
 - llvmlite=0.13.0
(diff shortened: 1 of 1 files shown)


dhimmel:
Here's a screenshot of how snorkel's `python-package-requirement.txt` [changes](https://github.com/HazyResearch/snorkel/compare/e49ecb2047327173ed56b2f85d4a71bdc4487bbc...c8a8d2377fe0f5d71cc078cd5368ac6f3b33a433#diff-1ea4f291a781edb002629c1df453f251) with the submodule update.

python-package-requirement.txt

As far as I can tell, the only change is that `theano` has been replaced by `tensorflow`. @ajratner will all uses need `tensorflow`? @danich1 so I think we should remove `theano` (check nothing breaks) and potentially add `tensorflow`, although we can hold off installing that until we need it, if you would like. Let's wait to hear from @ajratner.

ajratner:
@dhimmel Yes we moved all the default end extraction models to TF in this release, so this would be the right way to go!

dhimmel:
@danich1 let's see if [tensorflow installs](https://www.tensorflow.org/get_started/os_setup#anaconda_installation) with `pip`. Otherwise, we can use conda and add the `conda-forge` channel to our `environment.yml`.

danich1:
Added the tensorflow install line. Pip installs it like a charm!!!

dhimmel (approved):

PR #10, opened 2017-01-28
Extract epilepsy-associated genes trained using Hetionet v1.0 relationships

danich1:
Closes #7 
Closes #2 

This pull request contains working jupyter code that will parse the epilepsy abstracts, run a few labeling functions and run a generative model (Naive Bayes) to model the labeling functions. @dhimmel Let me know what you think. **Side Note:** I intentionally left out the data for the notebooks because the database is about ~282.2 mb

PR #11, opened 2017-02-01
Convert bioconcepts2pubtator_offsets to bioC format

danich1:
This pull request is designed for converting [Pubtator's abstracts annotations](ftp://ftp.ncbi.nlm.nih.gov/pub/lu/PubTator/) into Pubtator's xml format and to grab annotations in tsv format for our custom Tagger class in #10. 

There is a lot going on in pubtator_to_xml.py but the central idea is to take input like this:
>27117697|t|Quality assessment of studies comparing percutaneous ablative treatments in hepatocellular carcinoma.
27117697|a|
27117697	76	100	hepatocellular carcinoma	Disease	MESH:D006528

And convert it into this format:
> \<collection\>
\<source\>PubTator\<\/source\>
\<key\>PubTator.key\</key\>
\<document\>
\<id\>27117697\</id\>
\<passage\>
\<infon key="type"\>title\</infon\>
\<offset\>0\</offset\>
\<text\>Quality assessment of studies comparing percutaneous ablative treatments in hepatocellular carcinoma.\</text\>
\<annotation id="0"\>
\<infon key="type"\> Disease \</infon\>
\<location length="5" offset="6"\>\</location\>
\<text\>hepatocellular carcinoma\</text\>
\<infon key="MESH"\>D006528\</infon\>
\</annotation\>
\</passage\>
\<passage\>
\<infon key="type"\>abstract\</infon\>
\<offset\>101\</offset\>
\<text\>\</text\>
\</passage\>
\</document\>
\</collection\>


dhimmel:
@danich1 for the two file formats you mention above, can you comment on where there are from and where they are consumed? Are these formats that we're choosing, or that we're obligated to use since they're related to PubTator/Snorkel?

environment.yml
```diff
@@ -22,4 +22,5 @@ dependencies:
 - sqlalchemy=1.1.4
 - pip:
     - git+https://github.com/HazyResearch/numbskull@40ac1af20538c17b4726c963c69adcb81314efa5
-    - tensorflow==0.12.1
\ No newline at end of file
+    - tensorflow==0.12.1
+    - tqdm==4.11.2
```
dhimmel: Let's add a newline at the end, so future diffs are nicer.

Data/pickle_tags.py
```diff
@@ -0,0 +1,36 @@
+from lxml import etree as ET
+import tqdm
```
dhimmel: Let's separate builtin imports above (here just `sys`).

Data/pickle_tags.py
```diff
@@ -0,0 +1,36 @@
...
+
+def convert(xml_file, output_file):
+    """ Extract the annotations from pubtator xml formatted file
+
+    Keywords arguments:
+    xml_file -- The path to the xml data file
+    output_file -- the path to output the formatted data
+    """
+    with open(output_file, "w") as g:
+        g.write("Document\tType\tID\tOffset\tEnd")
```
dhimmel: Try to never hardcode file parsing. Use buitlin `csv` module instead. This way you specify the delimiter (`\t`) once, and pass the `csv.writer` lists for each row. Less error prone when you let the machine do the repetitive work.

Data/pickle_tags.py
```diff
@@ -0,0 +1,36 @@
...
+
+            # cycle through all the annotation tags contained within document tag
+            for annotation in document.iter('annotation'):
+                # not all annotations will contain an ID
+                if len(annotation) > 3:
+                    ant_type = annotation[0]
+                    ant_id = annotation[3]
+                    offset = int(annotation[1].attrib['offset'])
+                    end = offset + int(annotation[1].attrib['length'])
+                    g.write("%s\t%s\t%s\t%s\n" % (pubmed_id, ant_type, ant_id, offset, end))
```
dhimmel: Let's avoid the deprecated % string format method and use `.format` instead. This specific line will not matter once you switch to the csv module

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    term_tags = set()
+    with open(output_file, 'w') as g:
+        # Header for pubtator's xml format
+        g.write("<collection>\n<source>Pubtator</source>\n")
+        # read from gunzip file
+        with gzip.open(input_file, "rb") as f:
+
+            for line in tqdm.tqdm(f):
+                # Convert "illegal chracters" (i.e. < > &) in the main text
```
dhimmel: What defines these as illegal characters? I'm hoping there is an automated encoder/decoder, so we don't have to hardcode these conversions.

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,152 @@
+import gzip
+
+import tqdm
+
+
+def generate_annotation(file_object, tagged_term, index):
```
dhimmel: Would love if we could use an existing xml library so you don't have to hardcode the XML creation. Similar to using the `csv` module to write TSVs.

danich1:
The first file format comes directly from the [Pubtator's](https://www.ncbi.nlm.nih.gov/CBBresearch/Lu/Demo/PubTator/) ftp site. I downloaded the ~~``bioconcepts2pubtator.gz``~~ ``bioconcepts2pubtator_offsets.gz`` file, which consists of all the Bioconcept's annotations for each pubmed abstract that pubtator has stored. We are forced to deal with the raw data in this format, because I don't believe there is a way to get around this.

In regards to the second file format, this is referenced in [Pubtator's Tutorial](https://www.ncbi.nlm.nih.gov/CBBresearch/Lu/Demo/PubTator/tutorial/#ExportannotationinPubTator) and it looks like the format is [BioC](http://bioc.sourceforge.net/), which is an xml format specifically designed to share text data along with its corresponding annotations. We are free to use other formats for keeping track of these annotations; however, I believe that using BioC's xml format is the way to go because it is the easiest to work with. (AKA snorkel has their cdr tutorial which uses this specific format.)

Data/pubtator_to_xml.py
danich1: Are you talking about lxml module? If so, I did use that before this pull request; however, it didn't scale and ended up crashing my computer.

Data/pubtator_to_xml.py
danich1: This is mainly for my reference, when I fixed this section. [Solution](http://stackoverflow.com/questions/10993692/python-convert-to-html-special-characters)

dhimmel:
`bioconcepts2pubtator` looks like a TSV to me according to the sample (`ftp://ftp.ncbi.nlm.nih.gov/pub/lu/PubTator/bioconcepts2pubtator.sample`). For example, the head of the sample is:

```
PMID	Type	Database	Identifier	Mentions	Resource
10	Chemical	MESH|CHEBI	MESH:D004074	Digitoxin	MeSH|tmChem
1000006	Chemical	MESH|CHEBI	MESH:D004958	estradiol	MeSH|tmChem
1000006	Chemical	MESH|CHEBI	MESH:D011374	progesterone	MeSH|tmChem
1000007	Chemical	MESH|CHEBI	MESH:D004967	estrogen	MeSH
```

Nice to see that BioC is supposedly a standardized format. It looks like they have a [python package](https://github.com/2mh/PyBioC) for reading and writing BioC files -- however, it's not clear from the GitHub that's it's a reliable package. However, maybe worth a try.

dhimmel:
@danich1 I think you may have been referring to `bioconcepts2pubtator_offsets` (`ftp://ftp.ncbi.nlm.nih.gov/pub/lu/PubTator/bioconcepts2pubtator_offsets.sample`). The head of this file's sample for the first article is:

```
27037803|t|A case of cervical esophageal duplication cyst in a newborn infant.
27037803|a|Esophageal duplication cyst is a rare congenital anomaly resulting from a foregut budding error during the fourth to sixth week of embryonic development. Cervical esophageal duplication cysts are very rare and may cause respiratory distress in infancy. A full-term newborn girl who was born by normal delivery was transferred to our hospital because of swelling of the right anterior neck since birth. Cervical ultrasonography showed a 40   *   24   *   33  mm simple cyst on the right neck. Tracheal intubation was required at 2  weeks of age because of worsening external compression of the trachea. Fine-needle aspiration cytology revealed the existence of ciliated epithelium. At 1  month of age, exploration was performed through a transverse neck incision. The cyst had a layer of muscle connected to the lateral wall of the esophagus. Histopathological diagnosis was a cervical esophageal duplication cyst. We describe the clinical features of infantile cervical esophageal duplication cysts based on our experience of this rare disease in a neonate, along with a review of 19 cases previously reported in literature.
27037803	106	124	congenital anomaly	Disease	MESH:D000013
27037803	288	308	respiratory distress	Disease	MESH:D012128
27037803	452	456	neck	Disease	MESH:D006258
27037803	554	558	neck	Disease	MESH:D006258
27037803	661	668	trachea	Disease	MESH:D055090
27037803	816	820	neck	Disease	MESH:D006258
27037803	1099	1111	rare disease	Disease	MESH:D035583
27037803	60	66	infant	Species	9606
27037803	341	345	girl	Species	9606
```

This file has the full text.

dhimmel:
@ajratner, what is the best format of tagged literature for snorkel to consume?

Should we be converting the full pubtator export to BioC or do you recommend something else?

Data/pubtator_to_xml.py
dhimmel: We'll want to be defining the XML objects structure and letting a library do the rest ([example](http://stackoverflow.com/a/3605831/4651668)). A good writer should automatically deal the [encoding issues](https://github.com/greenelab/snorkeling/pull/11/files#r98961233). Options are PyBioC, lxml, or the builtin xml.

Regarding the scaling, we can figure out a way to do this iteratively (one article at a time).

dhimmel: This fork of PyBioC may be better https://github.com/en-dash/PyBioC. PyBioC looks unmaintained. @en-dash can you confirm whether https://github.com/en-dash/PyBioC/commit/a21bd181ce3b89de791db4bb7db6ca1c48167648 is better and production ready?

Data/pubtator_to_xml.py
dhimmel: [Here are some escaping methods](https://wiki.python.org/moin/EscapingHtml). In python 3, there's `html.escape`, but in two you're stuck with [`cgi.escape`](https://docs.python.org/2/library/cgi.html#cgi.escape) with `quote=True`. But I also am hopeful that we won't have to do this manually at all.

Data/pubtator_to_xml.py
dhimmel: Here's a BioC example from an [unmerged pull request](https://github.com/2mh/PyBioC/pull/2/files#diff-04c6e90faac2675aa89e2176d2eec7d8R28) to "fix broken example".

lfurrer: I'm pretty sure it has less bugs than the original version, but "production ready" is a big claim... I have succesfully used it as an importer and exporter tool, but I haven't done exhaustive testing.
Scaling up was an issue for me, too: When serialising a whole Medline chunk as a collection (30k abstracts plus annotations), PyBioC proved to be a serious bottleneck both in terms of memory consumption and processing time. The situation improved with the iterative serialisation method I added (one article at a time), but the corresponding method on the input size (using `iterparse()`) is still missing, I think.

lfurrer: Btw.: The commits of the other two open pull requests are all included in my fork.

dhimmel: @en-dash cool, sounds like [`en-dash/PyBioC`](https://github.com/en-dash/PyBioC) is the superior choice.

What's the syntax for the iterative writing of documents?

lfurrer: It's `BioCWriter.iterfragments()`. Iterates over chunks of serialised XML fragments:
```python
with open(outfile, 'wb') as f:  # XML is a binary format
    for frag in writer.iterfragments():
        f.write(frag)
```

Data/pickle_tags.py
```diff
@@ -0,0 +1,47 @@
+import csv
+from lxml import etree as ET
```
dhimmel: lxml should go with tqdm as non-bultin imports. 1 line separating this group from builtins.

Data/pickle_tags.py
```diff
@@ -0,0 +1,47 @@
...
+    Keywords arguments:
+    xml_file -- The path to the xml data file
+    output_file -- the path to output the formatted data
+    """
+    with open(output_file, "w") as csvfile:
+        fieldnames = ['Document', 'Type', 'ID', 'Offset', 'End']
+        writer = csv.DictWriter(csvfile, fieldnames=fieldnames)
+        writer.writeheader()
+
+        for event, document in tqdm.tqdm(ET.iterparse(xml_file, tag="document")):
```
dhimmel: For better readability, can we pull `ET.iterparse(xml_file, tag="document")` out of the for loop line and assign it to a variable above?

Data/pickle_tags.py
```diff
@@ -0,0 +1,47 @@
...
+def extract_annotations(xml_file, output_file):
+    """ Extract the annotations from pubtator xml formatted file
+
+    Keywords arguments:
+    xml_file -- The path to the xml data file
+    output_file -- the path to output the formatted data
+    """
+    with open(output_file, "w") as csvfile:
+        fieldnames = ['Document', 'Type', 'ID', 'Offset', 'End']
+        writer = csv.DictWriter(csvfile, fieldnames=fieldnames)
```
dhimmel: Great use of `csv.DictWriter`. Referring to things by name (rather than position) is the least error prone.

Data/pickle_tags.py
```diff
@@ -0,0 +1,47 @@
...
+        writer.writeheader()
+
+        for event, document in tqdm.tqdm(ET.iterparse(xml_file, tag="document")):
+            pubmed_id = document[0].text
+
+            # cycle through all the annotation tags contained within document tag
+            for annotation in document.iter('annotation'):
+                # not all annotations will contain an ID
+                if len(annotation) > 3:
```
dhimmel: To avoid indentation hell, consider:

```python
if len(annotation) <= 3:
    continue
```

Data/pickle_tags.py
```diff
@@ -0,0 +1,47 @@
...
+                    for infon in annotation.iter('infon'):
+                        if infon.attrib["key"] == "type":
+                            ant_type = infon.text
+                        else:
```
dhimmel: Can we be more explicit here rather than just else? Does the text always use the same XML key?

This can help understand the code from reading it.

Data/pickle_tags.py
```diff
@@ -0,0 +1,47 @@
...
+                            ant_id = infon.text
+
+                    for location in annotation.iter('location'):
+                        offset = int(location.attrib['offset'])
+                        end = offset + int(location.attrib['length'])
+
+                    writer.writerow({'Document': pubmed_id, 'Type': ant_type, 'ID': ant_id, 'Offset': offset, 'End': end})
```
dhimmel: Would it make sense to define a dict `row` right after the continue above and just write directly to it?

Data/pickle_tags.py
```diff
@@ -0,0 +1,47 @@
...
+            for annotation in document.iter('annotation'):
+                # not all annotations will contain an ID
+                if len(annotation) > 3:
+                    for infon in annotation.iter('infon'):
+                        if infon.attrib["key"] == "type":
+                            ant_type = infon.text
+                        else:
+                            ant_id = infon.text
+
+                    for location in annotation.iter('location'):
```
dhimmel: Are there ever multiple locations, i.e. `annotation.iter('location')` has multiple elements? If so, you're only taking last one. If always len 1, let's use:

```python
location, = annotation.iter('location')
```

Which will throw error if not len 1

Data/pickle_tags.py
```diff
@@ -0,0 +1,47 @@
...
+                        end = offset + int(location.attrib['length'])
+
+                    writer.writerow({'Document': pubmed_id, 'Type': ant_type, 'ID': ant_id, 'Offset': offset, 'End': end})
+
+            # prevent memory overload
+            document.clear()
+
+
+# Main
+xml_file = sys.argv[1]  # path to file im using now /home/davidnicholson/Documents/snorkeling/data/Epilepsy/epilepsy_data.xml
```
dhimmel: Consider argparse and using named arguments.

Data/pickle_tags.py
```diff
@@ -0,0 +1,47 @@
+import csv
```
dhimmel: I didn't see any pickling in this script. What does `pickle_tags.py` refer to?

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,161 @@
...
+    location = BioCLocation()
+    location.offset = str(tag[1])
+    location.length = str(len(tag[3]))
+    annt.locations.append(location)
+    annt.text = tag[3]
+    return annt
+
+
+def convert_pubtator(input_file, output_file):
+    """Convert pubtators annotation list to XML
```
dhimmel: Specify specific filename to avoid confusion.

dhimmel: Should doc specify that conversion is to BioC XML?

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,161 @@
...
+    collection.date = time.strftime("%Y/%m/%d")
+    collection.source = "Pubtator"
+    collection.key = "Pubtator.key"
+
+    with open(output_file, 'wb') as g:
+        term_tags = set()
+
+        # Have to manually do this because hangs otherwise
+        shell = writer.tostring('UTF-8')
+        tail = u'</collection>\n'
```
dhimmel: May be easier to do `from __future__ import string_literals` so you don't have to specify unicode. We can discuss.

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,161 @@
...
+    # Set up BioCWriter to write specifically Pubtator
+    # Can change to incorporate other sources besides pubtator
+    writer = BioCWriter()
+    writer.collection = BioCCollection()
+    collection = writer.collection
+    collection.date = time.strftime("%Y/%m/%d")
+    collection.source = "Pubtator"
+    collection.key = "Pubtator.key"
+
+    with open(output_file, 'wb') as g:
```
dhimmel, line 161: If we have encoding issues, we may need:

```
codecs.open(output_file, 'w', 'utf-8')
```

We can also consider `from eight import *` (https://github.com/kislyuk/eight). But I'm hesitant to recommend `eight` since I don't know how production ready it is. If it works correctly, it will make Python 2 behave like Python 3.

dhimmel: No action needed on this now

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,161 @@
...
+        term_tags = set()
+
+        # Have to manually do this because hangs otherwise
+        shell = writer.tostring('UTF-8')
+        tail = u'</collection>\n'
+        head = shell[:-len(tail)]
```
dhimmel: What is this select doing?

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,161 @@
...
+        g.write(head)
+
+        # read from gunzip file
+        with gzip.open(input_file, "rb") as f:
+            for line in tqdm.tqdm(f):
+                # Convert "illegal chracters" (i.e. < > &) in the main text
+                # into html entities
+                line = cgi.escape(line.strip()).encode("ascii", "xmlcharreplace")
```
dhimmel: Sure we don't want `utf-8`?

Should xmlcharreplace be xmlcharrefreplace?

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,161 @@
...
+
+    location = BioCLocation()
+    location.offset = str(tag[1])
+    location.length = str(len(tag[3]))
+    annt.locations.append(location)
+    annt.text = tag[3]
+    return annt
+
+
+def convert_pubtator(input_file, output_file):
```
dhimmel: Let's make a distinct function for reading `bioconcepts2pubtator_offsets` that yields articles.

dhimmel: How about defining a function like

```python
lines = list()
for line in read_file:
   line = line.rstrip()
   if line:
       lines.append(line)
    else:
       # Process title and abstact lines
       # Pass remaining lines through a csv.DictReader
       yield article
       lines = list()
```

Note indentation above is wrong.

dhimmel: Happy to chat about this since it's complex.

dhimmel (commented):
Haven't got all the way through. Main issue is splitting out the `bioconcepts2pubtator_offsets` parsing into a  dedicated function. Would be ideal if the BioC writer function was modular, so could apply to things other than converting `bioconcepts2pubtator_offsets`.

Data/pickle_tags.py
danich1: Originally it was for writing tags to a pickled file, but now that we changed the functionality of this file I renamed it.

Data/pubtator_to_xml.py
danich1: This part is writing the head and tail of the BioC xml file. It is from the write fragments function mentioned [here](https://github.com/greenelab/snorkeling/pull/11#discussion_r99014162). I can't use that function, since it involves me building the entire object in memory.

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,180 @@
...
+                # set up the csv reader
+
+                annts = csv.DictReader(file_lines[2:], fieldnames=['Document', 'Start', 'End', 'Term', 'Type', 'ID'], delimiter=str("\t"))
+                sorted_annts = sorted(annts, key=lambda x: x["Start"])
+                article["Title_Annot"] = filter(lambda x: x["Start"] < title_len, sorted_annts)
+                article["Abstract_Annot"] = filter(lambda x: x["Start"] > title_len, sorted_annts)
+
+                yield article
+                file_lines = list()
+
```
dhimmel: Let's add `assert not file_lines` outside loop to make sure we're not missing anything

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,180 @@
...
+    location = BioCLocation()
+    location.offset = str(tag["Start"])
+    location.length = str(len(tag["Term"]))
+    annt.locations.append(location)
+    annt.text = tag["Term"]
+    return annt
+
+
+def bioconcepts2pubtator_offsets(input_file):
```
dhimmel: Docstring. Specifically, what is the structure of `article` that gets yielded?

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,180 @@
...
+                title_len = len(title_heading[1])
+                article["Title"] = [title_heading[0], title_heading[1]]
+
+                # abstract
+                abstract_heading = file_lines[1]
+                article["Abstract"] = [abstract_heading[0], title_len, abstract_heading[1]]
+
+                # set up the csv reader
+
+                annts = csv.DictReader(file_lines[2:], fieldnames=['Document', 'Start', 'End', 'Term', 'Type', 'ID'], delimiter=str("\t"))
```
dhimmel: Did unicode_literals make you have to put `str()` around `"\t"`.

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,180 @@
...
+    collection.source = "Pubtator"
+    collection.key = "Pubtator.key"
+
+    with open(output_file, 'wb') as g:
+        term_tags = set()
+
+        # Have to manually do this because hangs otherwise
+        # Write the head of the xml file
+        xml_header = writer.tostring('UTF-8')
+        xml_end = u'</collection>\n'
```
dhimmel: Don't need the `u` if doing `from __future__ import unicode_literals`.

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,180 @@
...
+    # Can change to incorporate other sources besides pubtator
+    writer = BioCWriter()
+    writer.collection = BioCCollection()
+    collection = writer.collection
+    collection.date = time.strftime("%Y/%m/%d")
+    collection.source = "Pubtator"
+    collection.key = "Pubtator.key"
+
+    with open(output_file, 'wb') as g:
+        term_tags = set()
```
dhimmel: Is `term_tags` ever used?

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,180 @@
...
+            document.add_passage(title_passage)
+            document.add_passage(abstract_passage)
+
+            step_parent = E('collection')
+            writer._build_documents([document], step_parent)
+            g.write(tostring(step_parent[0], pretty_print=True))
+            step_parent.clear()
+
+        # Write the closing tag of the xml document
+        g.write(xml_tail)
```
dhimmel, line 204: Is `xml_tail` defined?

Data/extract_tags.py
```diff
@@ -0,0 +1,62 @@
+import argparse
+import csv
+import sys
+
+from lxml import etree as ET
+import tqdm
+
+
+def extract_annotations(xml_file, output_file):
+    """ Extract the annotations from pubtator xml formatted file
```
dhimmel, line 10: Docstring needs to specify the format of the output data... TSV it looks like?

dhimmel (commented):
Coming along nicely. I suggest testing the code on a small amount of the `bioconcepts2pubtator_offsets` file (you may already be doing this). Perhaps we can make an `example-data` with the small datasets at each stage: pubtator, bioc, and TSV.

Data/pubtator_to_xml.py
danich1: yes because csv module doesn't let me use unicode for delimiters

Data/pubtator_to_xml.py
danich1: Assertion failed, so I fixed this problem by adding a checker if statement

dhimmel: DRY!

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,206 @@
...
+
+    location = BioCLocation()
+    location.offset = str(tag["Start"])
+    location.length = str(len(tag["Term"]))
+    annt.locations.append(location)
+    annt.text = tag["Term"]
+    return annt
+
+
+def article_generator(file_lines):
```
dhimmel: Let's rename to something like `pubtator_stanza_to_article`. Generator is what a function that yields returns.

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,206 @@
...
+    """Article Generator
+
+    Returns an article that is a dictionary with the following keywords:
+    Title- an array of the document id and the title string
+    Abstract-  an array of document id, title offset, and abstract string
+    Title_Annot- A filtered array of tags specific to the title
```
dhimmel: These are lists not arrays (python nomenclature). Note that in PY3, the filter would return a generator, so you'd have to wrap in list (not problem here)

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,206 @@
...
+    annt.locations.append(location)
+    annt.text = tag["Term"]
+    return annt
+
+
+def article_generator(file_lines):
+    """Article Generator
+
+    Returns an article that is a dictionary with the following keywords:
+    Title- an array of the document id and the title string
```
dhimmel: Let's make `Title` a string of the just the title. Then you can make a `Pubmed_id` field.

dhimmel: Will come to discuss this.

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,206 @@
...
+
+    Keywords:
+    input_file - the name of the bioconcepts2putator_offset file (obtained from pubtator's ftp site: ftp://ftp.ncbi.nlm.nih.gov/pub/lu/PubTator/)
+    """
+    file_lines = list()
+    with open(input_file, "rb") as f:
+        for line in f:
+            # Convert "illegal chracters" (i.e. < > &) in the main text
+            # into html entities
+            line = cgi.escape(line.rstrip()).encode("ascii", "xmlcharrefreplace")
```
dhimmel: This shouldn't be here, this function is prior to any output encoding.

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,206 @@
...
+
+        # Have to manually do this because hangs otherwise
+        # Write the head of the xml file
+        xml_header = writer.tostring('UTF-8')
+        xml_tail = '</collection>\n'
+        xml_head = xml_header[:-len(xml_tail)]
+        g.write(xml_head)
+
+        # Write each article in BioC format
+        for article in tqdm.tqdm(bioconcepts2pubtator_offsets(input_file)):
```
dhimmel: style: lot's going on in this line. Consider 

```python
article_generator = bioconcepts2pubtator_offsets(input_file)
for article in tqdm.tqdm(article_generator):
```

environment.yml
```diff
@@ -22,4 +22,7 @@ dependencies:
 - sqlalchemy=1.1.4
 - pip:
     - git+https://github.com/HazyResearch/numbskull@40ac1af20538c17b4726c963c69adcb81314efa5
-    - tensorflow==0.12.1
\ No newline at end of file
+    - git+https://github.com/en-dash/PyBioC@a21bd181ce3b89de791db4bb7db6ca1c48167648
+    - tensorflow==0.12.1
+    - tqdm==4.11.2
+
```
dhimmel: Add final newline

Data/pubtator_to_xml.py
danich1: I'm not sure I follow this comment. Why do we need to remove it?

dhimmel: This function yields articles that are python representations of articles & tags. XML encoded should occur when writing to XML, not prior. Therefore, this should be moved to wherever article dictionaries are being converted to XML. Also, I'd think the BioC package should do this encoding for you?

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,215 @@
...
+        # Write the head of the xml file
+        xml_header = writer.tostring('UTF-8')
+        xml_tail = '</collection>\n'
+        xml_head = xml_header[:-len(xml_tail)]
+        g.write(xml_head)
+
+        article_generator = bioconcepts2pubtator_offsets(input_file)
+        # Write each article in BioC format
+        for article in tqdm.tqdm(article_generator):
+            id_index = 0
```
dhimmel: Let's move `id_index` to right above the for loops where it's used. Also `id_index = id_index + 1` is the same as `id_index += 1` if you want the brevity.

Data/pubtator_to_xml.py
dhimmel: Fixed.

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,215 @@
...
+
+            step_parent = E('collection')
+            writer._build_documents([document], step_parent)
+            g.write(tostring(step_parent[0], pretty_print=True))
+            step_parent.clear()
+
+        # Write the closing tag of the xml document
+        g.write(xml_tail)
+
+# Main
```
dhimmel, line 206:
```python
if __name__ == '__main__':
    # Rest of script goes here, so you can import functions from this file without running it as a script.
```

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,215 @@
...
+parser = argparse.ArgumentParser(description='Extracts the annotations from the BioC xml format')
+parser.add_argument("--documents", nargs=1, help="File path pointing to input file.")
```
dhimmel: `args.documents` shouldn't be a list, so let's remove `nargs=1`.

Data/pubtator_to_xml.py
```diff
@@ -0,0 +1,215 @@
...
+parser.add_argument("--output", nargs="?", help="File path for destination of output.")
```
dhimmel: Can also remove nargs here.

dhimmel: Maybe add `required=True` to both options.

dhimmel (commented):
This is coming along _really nicely_.

Just some finishing touches as well as the `examples` data directory.

Data/pubtator_to_xml.py
danich1: From looking at bioC xml format each annotation id is document dependent. In other words every document's first annotation has id of 0. If we want to remain strict to BioC xml format I argue to leave that line there.

Data/extract_tags.py
```diff
@@ -0,0 +1,69 @@
...
+    Offset - the character position where the term starts
+    End - the character position where the term ends
+
+    Keywords arguments:
+    xml_file -- The path to the xml data file
+    output_file -- the path to output the formatted data
+    """
+    with open(output_file, "w") as csvfile:
+        fieldnames = ['Document', 'Type', 'ID', 'Offset', 'End']
+        writer = csv.DictWriter(csvfile, fieldnames=fieldnames)
```
dhimmel, line 24: Let's use TSV rather than CSV... add `delimiter='\t'`

dhimmel: Also rename sample file to use a `.tsv` extension so GitHub does it's magic rendering.

dhimmel (commented):
In `Data/Example/` can we prepend the filenames with numbers to represent their order.

`1.sample_annotations.txt`
`2.sample_docs.xml`

I've found this really helps others understand what's going own (and yourself a year down the road).

environment.yml
```diff
@@ -21,5 +21,8 @@ dependencies:
 - scipy=0.18.1  
 - sqlalchemy=1.1.4
 - pip:
+    - future==0.16.0
```
dhimmel, line 24: anaconda should have `future`, no need for pip.

danich1: Hmm.. When I used the future line, I got an error. Maybe we should put it as a dependency instead?

dhimmel: What was the error? What do you mean "as a dependency"?

dhimmel:
PubTator parsing is now in a separate repo. https://github.com/greenelab/pubtator/pull/2 continues this pull request.

PR #15, opened 2017-06-28
Version upgrades to environment

danich1:
A very quick PR to update snorkel and make future PR not so large on the changes.


environment.yml
```diff
@@ -24,4 +24,6 @@ dependencies:
 - pip:
     - git+https://github.com/HazyResearch/numbskull@40ac1af20538c17b4726c963c69adcb81314efa5
     - neo4j-driver==1.1.0
+    - psycopg2==2.7.1
```
dhimmel: Get from anaconda https://anaconda.org/anaconda/psycopg2.

We always want to get a precompiled binary when possible from either the `anaconda` or `conda-forge` channel. This will make installation faster.

dhimmel (commented):
If we're doing updates, should we update some other parts of the environment as well? I think we want to update everything, unless snorkel requires a lower version

Here's how snorkel's requirements.txt has changed:

req-updates-snorkel

So let's lock lxml at 3.6.4, and upgrade rest?

Getting pandas to 0.20+ will be important for reading tables from compressed URLs.

environment.yml
```diff
@@ -1,27 +1,30 @@
...
+- scikit-learn=0.18.1
+- scipy=0.19.0 
+- sqlalchemy=1.1.11
 - pip:
     - git+https://github.com/HazyResearch/numbskull@40ac1af20538c17b4726c963c69adcb81314efa5
-    - neo4j-driver==1.1.0
-    - tensorflow==0.12.1
+    - neo4j-driver==1.3.1
+    - tensorflow==1.2.0
+    - tika==1.14.1
```
dhimmel: Can install [from conda-forge](https://anaconda.org/conda-forge/tika) like:

```
- conda-forge::tika=1.14.1
```

environment.yml
```diff
@@ -1,27 +1,30 @@
...
+    - tqdm==4.14.0
```
dhimmel: In anaconda https://anaconda.org/anaconda/tqdm

dhimmel (commented):
Looks like we need to also add `six` and `future`

+ https://anaconda.org/anaconda/six
+ https://anaconda.org/anaconda/future

These are probably installed as dependencies of things we specify, but let's list explicitly since they're new in the snorkel requirements file.

environment.yml
danich1: So looking into conda-forge's list of packages, there is a package for tensor flow but the version is a bit dated: 1.1.0 v 1.2.0. This leads to my question should I resort towards just dealing with conda-forge's package or let pip take the wheel?

dhimmel: Presumably since v1.2.0 was just [recently released](https://github.com/tensorflow/tensorflow/releases/tag/v1.2.0). Why don't we stick with pip and can switch to a precompiled binary when it's up to date.

dhimmel (approved):

Issue #20, opened 2017-07-11
Scaling Snorkeling To Handle Pubmed


David-Durst:
I've got the version of the Snorkeling project from the greenelab repo downloaded and running. See [1] at the bottom for a suggestion about that.

I can't run David's code [2] so I can't test the issue myself. However, I've read the code and have a few questions that will help me in my investigation.
 
1. What is the exact problem? Is it that CorpusParser.apply with the default implementation of XMLMultiDocPreparser runs out of memory when loading data from the file /home/davidnicholson/Documents/Data/pubmed_docs.xml? 

2. I heard something about a memory leak. Is it still thought that there is a memory leak? If so, why?

3. Why was chunking done? It appears that "corpus_parser.apply(xml_parser)", when using XMLMultiDocPreparser from snorkeling/All_Relationships/utils/bigdata_utils.py, should follow the scalable process of reading in one document from the XMLMultiDocPreparser, calling CorpusParserUDF.apply once, dereferencing that document so that it can be garbage collected, and then repeating. The parallel version of corpus_parser.apply should have each process follow this loop independently. 

# -- Notes --

[1] Not sure if it's just me, but in the future you guys might want to recommend that new developers also do these things:

Add conda-forge to their conda channels.
Command: conda config --add channels conda-forge

Install icu 56.1 from the conda-forge channel.
Command: conda install icu==56.1
 
I was having trouble loading the library lxml for Snorkel after sourcing the conda env due to a missing shared object from icu version 56. Installing this dependency fixed that.
 
[2] David's code refers to a Postgres database that I don't have and a path on David's computer that is not in the git repo, /home/davidnicholson/Documents/Data.

danich1:
Thanks for opening this issue. Plan to get everything (Fingers crossed) set up by Friday, where you are able to reproduce this issue effectively.

David-Durst:
Sounds good. Let me know if the project is setup for me to help further or if you solved the problem.

danich1:
So the problem looks it was resolved when I upgraded my ram from 32G to 64G. In light of this "magical" fix, I'm not sure if looking at the code will uncover any hidden issues; nonetheless, I will be getting the project up as soon as I can so you can still take a look if desired.

Issue #19, opened 2017-06-30
Conda forge breaks lxml 


danich1:
There is a bug when trying to rely on conda forge to install the python module lxml. Right now if one were to create the conda environment and import lxml.etree, the following error will occur:

>Traceback (most recent call last):
> File "<string>", line 1, in <module>
> ImportError: libicui18n.so.56: cannot open shared object file: No such file or directory

Therefore, I argue that we move lxml to be a pip dependency as installing lxml through that way fixes it.

dhimmel:
Lot's of related issues on GitHub:

+ https://github.com/chapmanb/bcbio-nextgen/issues/1878
+ https://github.com/poldracklab/mriqc/issues/371
+ https://github.com/conda-forge/geopandas-feedstock/issues/21

`lxml` via pip could be a good workaround. It looks like you may also be able to specify `icu=56.1`.

gwaybio:
Just encountered a similar situation and landed here searching for solutions. I was able to solve the conda forge error:

```
>>> import igraph
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/gway/anaconda3/envs/gea_community_detection/lib/python2.7/site-packages/igraph/__init__.py", line 34, in <module>
    from igraph._igraph import *
ImportError: libicui18n.so.58: cannot open shared object file: No such file or directory
```
by specifying `icu=58` in the environment.yml file in greenelab/GEA_Community_Detection#23

PR #21, opened 2017-07-18
Data Insertion Update

danich1:
Closes #18 and #19. This PR is for updating the repository with a version upgrade on the numbskull module and for uploading a notebook that allows one to insert data into a postgres database. Ideally, after this PR is finished work can be done on #20, if desired.


All_Relationships/1.data-loader.py
```diff
@@ -0,0 +1,232 @@
...
+#Grab the sentences!!!
+train_sens = set()
+dev_sens = set()
+test_sens = set()
+
+offset = 0
+category_index = 0
+has_docs = True
+#divde and insert into the database
+while has_docs:
```
dhimmel: What is changing in each iteration of `while has_docs`? I think we may be able to make this clearer.

All_Relationships/utils/__init__.py
```diff
@@ -0,0 +1 @@
+from utils import *
```
dhimmel: newline

All_Relationships/.ipynb_checkpoints/1.data-loader-checkpoint.ipynb
```diff
@@ -0,0 +1,394 @@
+{
```
dhimmel: This file is in an `.ipynb_checkpoints` directory. This is likely a temp file created by Jupyter and should be ignored?

Let's add the following to .gitignore

```
# Jupyter Notebook
.ipynb_checkpoints
```

All_Relationships/1.data-loader.py
```diff
@@ -0,0 +1,232 @@
...
+from sqlalchemy import func
+
+
+# # Parse the Pubmed Abstracts
+
+# The code below is designed to read and parse data gathered from pubtator. Pubtator outputs their annotated text in xml format, so that is the standard file format we are going to use. 
+
+# In[ ]:
+
+get_ipython().magic(u"time filter_df = pd.read_table('/home/danich1/Documents/pubtator/data/pubtator-hetnet-tags.tsv.xz')")
```
dhimmel: Can we use versioned GitHub URL like:

```
https://github.com/greenelab/pubtator/raw/631e86002e11c41cfcfb0043e60b84ab321bdae3/data/pubtator-hetnet-tags.tsv.xz
```

All_Relationships/1.data-loader.ipynb
```diff
@@ -0,0 +1,394 @@
+{
```
dhimmel, line 1: The output of this notebook is blank. In general, I like to commit the executed notebook, so we can see the output. I'll usually do "Restart & Run All" to make sure the notebook is fresh... then commit. Does this notebook take a long time to run? If so, it may not be worth rerunning.

All_Relationships/1.data-loader.py
```diff
@@ -0,0 +1,232 @@
...
+            
+            if category == 0:
+                train_sens.add(s)
+            elif category == 1:
+                dev_sens.add(s)
+            else:
+                test_sens.add(s)
+
+    if has_docs:
+        # Disease-Gene Edge
```
dhimmel: Seems repetetive. Should use for loop like:

```
for edges in dge, gge, cge, cde
```

All_Relationships/1.data-loader.ipynb
danich1: Yeah can't rerun, because it will take 3+ days. Plus it will generate a lot of output, so I'd refrain from going this route.

dhimmel (approved):

PR #23, opened 2017-08-31
Data gen training

danich1:
This PR is for updating the snorkeling repo with the data-labeler and the data-gen model notebooks. Basically these two notebooks cover the whole data generation and generative model training section of this project. Unfortunately, this is one of two big repos. (Sacrificed updating frequency to get things running). Let me know what you think @dhimmel. Got a lot of tacos coming your way after these two repos are done.


All_Relationships/2.data-labeler.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+# In[ ]:
+
+
+if edge_type == "dg":
+    DiseaseGene = candidate_subclass('DiseaseGene', ['Disease', 'Gene'])
+    edge = "disease_gene"
+elif edge_type == "gg":
+    GeneGene = candidate_subclass('GeneGene', ['Gene1', 'Gene2'])
```
dhimmel, line 64: What is the motivation here to do `['Gene1', 'Gene2']` rather than `['Gene', 'Gene']`?

All_Relationships/2.data-labeler.py
```diff
@@ -0,0 +1,221 @@
...
+
+# # Label Functions
+
+# Here is one of the fundamental part of this project. Below are the label functions that are used to give a candidate a label of 1,0 or -1 which corresponds to correct label, unknown label and incorrection label. The goal here is to develop functions that can label accurately label as many candidates as possible. This idea comes from the [data programming paradigm](https://papers.nips.cc/paper/6523-data-programming-creating-large-training-sets-quickly), where the goal is to be able to create labels that machine learning algorithms can use for accurate classification.  
+
+# In[ ]:
+
+
+if edge_type == "dg":
+    from utils.disease_gene_lf import *
```
dhimmel, line 108: best to avoid `import *`, which makes it difficult for readers to know where imports are coming from. Would it be problematic to just list the functions being imported?

All_Relationships/3.data-gen-model.py
```diff
@@ -0,0 +1,136 @@
+
+# coding: utf-8
+
+# # MUST RUN AT THE START OF EVERYTHING
```
dhimmel, line 4: Is this in reference to the first cell only? It's formatted as the title for the entire notebook. I'd recommend titling the notebook by what steps it performs in the overall analysis pipeline.

dhimmel (commented):
> Unfortunately, this is one of two big repos

I assume you mean PRs not repos (:

I see lot's of merge commits... in future try to start PR from most recent upstream version and use rebase. It's a bit complicated, so happy to go through the git workflow together after we merge this.

I didn't notice any data files getting added besides `hetnet_dg_kb.csv`. What are the outputs of the two new notebooks? Is everything getting written the database, so you're not writing separate output files that we should track?

All_Relationships/2.data-labeler.py
danich1: So for the candidate subclass it needs to take two column names and they need to be unique; therefore, I appended the 1 and 2. I got an error when I used `['Gene', 'Gene']`.

danich1:
>I didn't notice any data files getting added besides hetnet_dg_kb.csv. What are the outputs of the two new notebooks? Is everything getting written the database, so you're not writing separate output files that we should track?

The output of the first two notebook is all inside the database. There is a section (in notebook 2) where a datafile is generated, but the purpose of that file is to load the data into the database. I didn't think we needed to track that information since all the data is already in the database and can be accessed through a database dump. 

I can add tracking to the ML models that get saved for each iteration, but I think it is better to add that tracking in the next PR (once LSTM is done training). If I should add the Gen model tracking in this PR lmk.

dhimmel (approved):
I can help you with the merge and getting your repo up to date afterwards.

PR #24, opened 2017-09-06
Adding Model Training/Analysis Notebooks

danich1:
PR round 2. This PR is adding the other two notebooks which are focused on training and analyzing discriminator models for candidate classification. Notebook 5 isn't fully completed yet as the LSTM hasn't finished running yet. Will finish that notebook in another PR.

### DIFF
--- a/All_Relationships/4.data-disc-model.py
+++ b/All_Relationships/4.data-disc-model.py
@@ -171,7 +171,7 @@ np.savetxt("pmacs/train_marginals", train_marginals)
 # In[ ]:
 
 
-get_ipython().run_cell_magic(u'time', u'', u"train_kwargs = {\n    'lr':         0.001,\n    'dim':        100,\n    'n_epochs':   10,\n    'dropout':    0.5,\n    'print_freq': 1,\n    'max_sentence_length': 1000,\n}\n\nlstm = reRNN(seed=100, n_threads=4)\n#lstm.train(train_cands, train_marginals[0:10], X_dev=dev_cands, Y_dev=L_dev[0:10], **train_kwargs)")
+get_ipython().run_cell_magic(u'time', u'', u'"""\ntrain_kwargs = {\n    \'lr\':         0.001,\n    \'dim\':        100,\n    \'n_epochs\':   10,\n    \'dropout\':    0.5,\n    \'print_freq\': 1,\n    \'max_sentence_length\': 1000,\n}\n"""\nlstm = reRNN(seed=100, n_threads=4)\n#lstm.train(train_cands, train_marginals[0:10], X_dev=dev_cands, Y_dev=L_dev[0:10], **train_kwargs)')
 
 
 # ### Write the Training data to an External File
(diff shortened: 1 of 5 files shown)


All_Relationships/4.data-disc-model.py
```diff
@@ -0,0 +1,234 @@
...
+
+# ### Write the Training data to an External File
+
+# In[ ]:
+
+
+import csv
+chunksize = 100000
+start = 0
+with open('pmacs/train_candidates_ends.csv', 'wb') as g:
```
dhimmel: The with statement can use multiple context managers in one statement:

```python
with open('pmacs/dev_candidates_ends.csv', 'wb') as g, open("pmacs/dev_candidates_offsets.csv", "wb") as f:
```

All_Relationships/4.data-disc-model.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+get_ipython().magic(u'time train_marginals = load_marginals(session, split=0)')
+np.savetxt("pmacs/train_marginals", train_marginals)
+
+
+# In[ ]:
+
+
+get_ipython().run_cell_magic(u'time', u'', u"train_kwargs = {\n    'lr':         0.001,\n    'dim':        100,\n    'n_epochs':   10,\n    'dropout':    0.5,\n    'print_freq': 1,\n    'max_sentence_length': 1000,\n}\n\nlstm = reRNN(seed=100, n_threads=4)\n#lstm.train(train_cands, train_marginals[0:10], X_dev=dev_cands, Y_dev=L_dev[0:10], **train_kwargs)")
```
dhimmel: `train_kwargs` is never used? The command that uses it is commented out... should this be commented out too?

All_Relationships/4.data-disc-model.py
```diff
@@ -0,0 +1,234 @@
...
+    for key in tqdm.tqdm(lstm.word_dict.d):
+        output.writerow({'Key':key, 'Value': lstm.word_dict.d[key]})
+
+
+# ### Save the Development Candidates to an External File
+
+# In[ ]:
+
+
+get_ipython().run_cell_magic(u'time', u'', u'dev_cands = session.query(DiseaseGene).filter(DiseaseGene.split == 1).order_by(DiseaseGene.id).all()')
```
dhimmel: There is a more readable whitespace convention for these piped commands:

```python
dev_cands = (
    session
    .query(DiseaseGene)
    .filter(DiseaseGene.split == 1)
    .order_by(DiseaseGene.id)
    .all()
)
```

All_Relationships/5.data-analysis.py
```diff
@@ -0,0 +1,292 @@
...
+
+model_marginals = pd.read_csv("disc_marginals.csv")
+top_pos_predict_model_marginals = model_marginals.sort_values("LR_Marginals", ascending=False).head(10)
+top_neg_predict_model_marginals = model_marginals.sort_values("LR_Marginals", ascending=True).head(10)
+
+
+# In[ ]:
+
+
+from collections import defaultdict
```
dhimmel: I think you want a `collections.Counter` here not a `collections.defaultdict`

dhimmel (commented):
`5.data-analysis.ipynb` is unexecuted? The point of the notebook is to see the output inline, so committing unexecuted notebooks is a bit counter-intuitive. Is the notebook not complete yet?

All_Relationships/4.data-disc-model.py
danich1: `train_kwargs` should be commented out since I don't use it here. The initialization of the LSTM object shouldn't be commented out since I need to use it to generate the training data for the cluster.

danich1:
I agree with your above comment; however, that notebook isn't completed yet. The LSTM is still running on the cluster. I can halt this PR and wait for that data if it makes more sense to do that.

dhimmel:
> The LSTM is still running on the cluster. I can halt this PR and wait for that data if it makes more sense to do that.

No that's fine. You can create another PR with the results.

dhimmel (approved):

PR #25, opened 2017-09-20
Documenting the First Run for snorkeling

danich1:
Documenting all the changes I made for this run, so I can refer back to the details next time. I also added files that I used for the cluster.


All_Relationships/Experiment 1/LSTM.py
```diff
@@ -0,0 +1,55 @@
+import csv
+import os
+
+import numpy as np
+import tqdm
+from snorkel.learning.disc_models.rnn import reRNN
+from snorkel.learning.disc_models.rnn.utils import SymbolTable
+
+
+def read_csv_file(filename, get_max_val=False):
```
dhimmel: This needs a docstring for what kind of CSV it reads and what it is doing with the max int operations

All_Relationships/Experiment 1/LSTM.py
```diff
@@ -0,0 +1,55 @@
...
+        input_file = csv.reader(f)
+        for row in tqdm.tqdm(input_file):
+            data.append(np.array(map(int, row)))
+            if get_max_val:
+                max_row = max(map(int, row)) 
+                val = max_row if max_row > val else val
+    return (np.array(data), (val+1)) if get_max_val else np.array(data)
+
+
+def read_word_dict(filename):
```
dhimmel: Docstrings. Here's an example:

```python
"""
Read a CSV into a dictionary using the Key column (as string) and Value column (as int).
"""
```

All_Relationships/Experiment 1/evaluate_lstm.py
```diff
@@ -0,0 +1,30 @@
...
+import os
+
+import numpy as np
+import pandas as pd
+import tqdm
+from snorkel.learning.disc_models.rnn import reRNN
+from snorkel.learning.disc_models.rnn.utils import SymbolTable
+
+
+def read_csv_file(filename, get_max_val=False):
```
dhimmel: Is this function redundant with the `read_csv_file` in `LSTM.py`

All_Relationships/Experiment 1/job_sub.sh
```diff
@@ -0,0 +1,3 @@
+source set_env.sh
+cd All_Relationships
+bsub -o ~/sub-output.out -e ~/sub-error.out -q gpu -R "select[ngpus>0] rusage [ngpus_shared=1]" -n 1 python LSTM.py
```
dhimmel: one nice trick for git diffs and readability is to split long shell commands onto multiple lines:

```sh
bsub \
  -o ~/sub-output.out \
  -e ~/sub-error.out -q gpu \
  -R "select[ngpus>0] rusage [ngpus_shared=1]" \
  -n 1 \
  python LSTM.py
```

dhimmel (commented):
I recommend against spaces in directory or file names unless there is a strong reason to include them. The issue with spaces is that they can require quoting or escaping in shell commands, which can add code complexity. Not a huge deal, just something to consider when naming directories or files. I generally use dashes in file names for spaces, unless they're python modules, in which case you gotta use underscores.

dhimmel:
The old files in the `Experiment 1` directory are still tracked... When renaming files, the best way is to use `git mv`. But now I think the best solution is `git rm -r "All_Relationships/Experiment 1"`

dhimmel (approved):

PR #27, opened 2017-10-18
Repo Clean up

danich1:
This PR is just to update the official repository with updated file names and a better version of running the LSTM on the pmacs cluster. Shouldn't need a lot of editing since I mainly just changed the name of files.


dhimmel (commented):
File name changes look good to me. Housekeeping like that is always good! Note that if you think you may exceed 10 notebooks, I'd suggest naming like `01.candidate-text-loader.ipynb`

I have some questions about `1a.stratify-candidates.ipynb` which will be easiest in person.

dhimmel (approved):

PR #30, opened 2017-11-09
Adding notebook for fisher exact test

### LINKED ISSUE #26: Assessing term co-occurrence across sentences
@danich1 has extracted a bunch of sentences that include both a gene and a disease. He's computing summary statistics for gene-disease pairs. One basic measure being the number of sentences with both the gene and disease. Now we want to compute the expected number of sentences, given the marginal frequency of the gene and disease.

Let's take an approach similar to [computing MEDLINE term co-occurrence](https://think-lab.github.io/d/67/#4). You will need for each gene disease pair to compute the values for a contingency table where:

+ a is the number of sentences containing both the gene and the disease (cooccurrence)
+ b is the number of sentences containing the gene but not the disease
+ c is the number of sentences containing the disease but not the gene
+ d is the number of sentences without either the gene or disease

We should limit ourselves to only sentences that contain a gene and a disease. You'll be able to compute the expected and the p-value from a fisher's exact test using code [like here](https://github.com/dhimmel/medline/blob/0c9e2905ccf8aae00af5217255826fe46cba3d30/cooccurrence.py#L56-L61).

The expected number of sentences is actually quite easy to compute. You just take the number of sentences with the gene * the number of sentences with the disease divided by the total number of sentences (only considering sentences with both a gene and disease).

danich1:
Closes #26. This PR is adding a notebook that calculates the fisher exact test on co-occurence terms. Let me know what you think.

### DIFF
--- a/All_Relationships/6.true_relationship_preparation.py
+++ b/All_Relationships/6.true_relationship_preparation.py
@@ -5,14 +5,14 @@
 
 # After predicting the co-occurence of candidates on the sentence level, the next step is to predict whether a candidate is a true relationship or just occured by chance. Through out this notebook the main events involved here are calculating summary statistics and obtaining the LSTM marginal probabilities.
 
-# In[14]:
+# In[ ]:
 
 
 get_ipython().magic(u'load_ext autoreload')
 get_ipython().magic(u'autoreload 2')
 get_ipython().magic(u'matplotlib inline')
 
-from collections import defaultdict
+from collections import Counter
 import csv
 import os
 
@@ -29,7 +29,7 @@ from sklearn.model_selection import RandomizedSearchCV
 import seaborn as sns
 
 
-# In[15]:
+# In[ ]:
 
 
 #Set up the environment
@@ -45,14 +45,14 @@ from snorkel import SnorkelSession
 session = SnorkelSession()
 
 
-# In[16]:
+# In[ ]:
 
 
 from snorkel.models import Candidate, candidate_subclass
 from snorkel.learning.disc_models.rnn import reRNN
 
 
-# In[17]:
+# In[ ]:
 
 
 DiseaseGene = candidate_subclass('DiseaseGene', ['Disease', 'Gene'])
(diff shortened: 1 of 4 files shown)


All_Relationships/6.true_relationship_preparation.py
```diff
@@ -0,0 +1,243 @@
...
+
+DiseaseGene = candidate_subclass('DiseaseGene', ['Disease', 'Gene'])
+
+
+# # Count the Number of Sentences for Each Candidate
+
+# In[ ]:
+
+
+get_ipython().run_cell_magic(u'time', u'', u'doc_counter = defaultdict(set)\nsentence_counter = defaultdict(set)\noffset = 0\nchunk_size = 1e5\n\nwhile True:\n    cands = session.query(DiseaseGene).limit(chunk_size).offset(offset).all()\n    \n    if not cands:\n        break\n        \n    for candidate in tqdm.tqdm(cands):\n        sentence_counter[(candidate.Disease_cid, candidate.Gene_cid)].add(hash(candidate[0].get_parent()))\n        doc_counter[(candidate.Disease_cid, candidate.Gene_cid)].add(hash(candidate[0].get_parent().document_id))\n    \n    offset+= chunk_size')
```
dhimmel: a `collections.Counter` is probably more appropriate here than a `defaultdict`?

All_Relationships/6.true_relationship_preparation.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+# In[ ]:
+
+
+candidate_df = pd.DataFrame(map(lambda x: [x[0], x[1], len(sentence_counter[x]), len(doc_counter[x])], sentence_counter), columns=["disease_id", "gene_id", "sentence_count", "doc_count"])
```
dhimmel: Add some newlines here so it's easier to read the command

All_Relationships/6.true_relationship_preparation.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+total = candidate_df["sentence_count"].sum()
+odds = []
+p_val = []
+expected = []
+lower_ci = []
+
+for disease, gene in tqdm.tqdm(zip(candidate_df["disease_id"],candidate_df["gene_id"])):
+    cond = (candidate_df["disease_id"] == disease) & (candidate_df["gene_id"] == gene)
```
dhimmel: Would `.query` be more readable for subsetting the df?

```python
cond_df = candidate_df.query("disease_id == @disease and @gene_id == gene")
```

All_Relationships/6.true_relationship_preparation.py
```diff
@@ -0,0 +1,243 @@
...
+dev_sentences_df["marginals"] = dev_marginals_df["RNN_marginals"].values
+test_sentences_df["marginals"] = test_marginals_df["RNN_10_Marginals"].values
+
+
+# # Combine Marginal Probabilities
+
+# In[30]:
+
+
+candidate_df = pd.read_csv("disease_gene_summary_stats.csv")
```
dhimmel: Why is `hetnet_labels` -1 and 1 rather than 1 / 0? I guess this is how snorkel encodes? For most applications 0/1 is more convenient

All_Relationships/6.true_relationship_preparation.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+candidate_df.head(10)
+
+
+# In[37]:
+
+
+candidate_marginals = (
+                        train_sentences_df[["disease_id", "gene_id", "marginals"]]
```
dhimmel: Indent here at four spaces... not this much.

All_Relationships/6.true_relationship_preparation.py
```diff
@@ -0,0 +1,243 @@
...
+candidate_df["quantile_eighty"] = quantile_eighty
+candidate_df["avg_marginal"] = avg_marginals
+
+
+# ## Save the data to a file
+
+# In[44]:
+
+
+candidate_df.to_csv("disease_gene_summary_stats.csv", index=False)
```
dhimmel: It's a bit odd to me how the data is read in with the results already in it. This sort of behavior indicates that the notebook isn't necessarily able to recreated using a sequential run.

All_Relationships/6.true_relationship_preparation.py
```diff
@@ -0,0 +1,243 @@
...
+    odds_ratio, p_value = fisher_exact(c_table, alternative='greater')
+    odds.append(odds_ratio)
+    p_val.append(p_value)
+    
+    
+    total_disease = candidate_df[candidate_df["disease_id"] == disease]["sentence_count"].sum()
+    total_gene = candidate_df[candidate_df["gene_id"] == gene]["sentence_count"].sum()
+    expected.append((total_gene * total_disease)/float(total))
+
+candidate_df["p_value"] = p_val
```
dhimmel: Let's store the `-log10(p_value)` and rename to `nlog10_p_value`. This value is easier to view and will be more useful for machine learning.

All_Relationships/6.true_relationship_preparation.py
dhimmel: This dataframe is really beautiful. Can we got a short summary of the non-self-explanatory columns. What is `avg_marginal` and `final_model_pred`?

danich1: It is -1 because snorkeling's encoding is 1 for true, 0 for unknown and -1 for false. Could change it to 0/1, but I think it will be okay with -1 and 1.

All_Relationships/6.true_relationship_preparation.py
danich1: I combined the true-relationship model's predictions (probabilities) within the same dataframe as the summary statistics. Hence the final_model_pred column. Looking back that was a bit excessive and am separating that into its own dataframe from here on.

All_Relationships/6.true_relationship_preparation.py
```diff
@@ -0,0 +1,238 @@
...
+    train_sentences_df[["disease_id", "gene_id", "marginals"]]
+    .append(dev_sentences_df[["disease_id", "gene_id", "marginals"]])
+    .append(test_sentences_df[["disease_id", "gene_id", "marginals"]])
+    )
+
+
+# In[ ]:
+
+
+group = candidate_marginals.groupby(["disease_id", "gene_id"])
```
dhimmel: This is repetitive. The preferred solution is to define a list with the desired quantiles. This would also be more modular, allowing easy update of the selected quantiles.

You can automatically generate the column names like:

```python
'quantile_{:.2f}'.format(quantile)
```

All_Relationships/6.true_relationship_preparation.py
```diff
@@ -0,0 +1,228 @@
...
+    dg_series = group.get_group((cand["disease_id"], cand["gene_id"]))
+    avg_marginals.append(dg_series["marginals"].mean())
+    quantile_data.append(map(lambda x: dg_series["marginals"].quantile(x), quantile_list))
+    if i == 10:
+        break
+print len(quantile_data[0])
+# Save the evidence into a dataframe
+candidate_df = pd.concat(
+    [
+        candidate_df.head(11),
```
dhimmel: What is this `.head(11)`

danich1: Removing that now. It was for testing the newly formatted code and I forgot to remove.

dhimmel (approved):
@danich1 squash merge at will!

Issue #28, opened 2017-11-01
Verifying Experimental Analysis Design


danich1:
I talked with @dhimmel  yesterday and we came up with a design for determining whether or not adding input from a deep learning model (LSTM) is beneficial for predicting relationships between Diseases and Genes. 

**Background**:

![project overview](https://user-images.githubusercontent.com/3505417/32292560-e8c27dea-bf16-11e7-943b-e0dd9f7e6764.png)

Within the image above we have all disease-gene pair mappings where some edges are mentioned in **pubmed abstracts** (noted by the black dashes) and majority of edges aren’t mentioned at all. The edges in green are considered true edges as they are currently contained in **hetnet v1** and the other edges (not highlighted) have the potential to be a true Disease-Gene relationship. We aim to classify each edge as either positive (True edge) or negative (False edge), under the hypothesis that using NLP and deep learning (Long short term memory networks or LSTM for short) will provide better accuracy than standard methods.

**Analysis Design**:
To test this hypothesis we plan to use the following design:

| Categories: | Prior | Co-occurrences | Natural Language Processing (NLP) |
| --- | --- | --- | --- |
| | 1 Model | 1 Model with sentences | 1 Model with Sentences | 
| | | 1 Model w/o Sentences | 1 Model w/o Sentences | 
| | Literature unaware | LSTM unaware | LSTM aware |

The prior category is where we plan to use a model to classify each disease-gene edge without using any information from biomedical literature (hence literature unaware). The co-occureence category is where we plan to use a model that combines the prior category model with information obtained from biomedical literature i.e. (expected number of sentences that mentions a given disease-gene pair, the p-value for each disease-gene edge, how many unique abstracts that mention a given disease-gene pair etc.) To note this model doesn’t use the LSTM and just relies on the features extracted from the literature itself. A challenge for this will be handling the edges that aren’t mentioned within the literature itself. (Model w/o Sentences) Lastly, the NLP category combines the other two models and adds input from a deep learning model (probability that a sentence is evidence for a true disease-gene relationship). We expect to see the NLP category model outperform the models from the other two categories.

**Challenges**: 
1. What is a fair prior model to use for this analysis?
2. What do we do about edges that are in hetnet, but aren’t mentioned in literature? How can we classify these edges?

dhimmel:
Great summary of our brainstorm @danich1!

> What is a fair prior model to use for this analysis?

[...]

PR #31, opened 2017-11-29
Adding prior probability calculation notebook

danich1:
Pretty quick PR. Adds code for calculating the prior probability by mimicking [this notebook](https://github.com/dhimmel/learn/blob/master/prior/1-prior.ipynb).

Refs https://github.com/greenelab/snorkeling/issues/28#issuecomment-341738328


All_Relationships/7.prior_probabilities_calculation.py
```diff
@@ -0,0 +1,149 @@
...
+
+disease_degree = dict(hetnet_df["disease_id"].value_counts())
+gene_degree = dict(hetnet_df["gene_id"].value_counts())
+
+
+# In[ ]:
+
+
+association_edge = defaultdict(set)
+association_row = list()
```
dhimmel: I do think you can be more clear by using plural for list variable names when appropriate. For example, any reason not to call this `association_rows`?

dhimmel (commented):
Looks good to me... but hard to tell without the output. Can this be run fresh and saved with the output? Unless there's output, there's no point in using notebooks over scripts.

dhimmel (approved):
Nice!

PR #34, opened 2017-12-20
Notebook Organization and Entity Level prediction

danich1:
This PR makes some organizational changes and adds notebooks for entity level predictions/feature generation.


All_Relationships/9.entity_prediction.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+n_iter = 100
+final_models = []
+
+lr = LogisticRegression()
+elastic_net = SGDClassifier(penalty='elasticnet', loss='log', random_state=100)
+
+lr_grid = {'C':np.linspace(1, 1000, num=100)}
+elastic_net_grid = {'alpha':np.linspace(1e-9,1, num=100), 'l1_ratio':np.linspace(0,1,num=10)}
```
dhimmel: Let's just pick an `l1_ratio`. Optimizing both `alpha` and `l1_ratio` is more likely to overfit. Going with ridge makes sense to me here (`l1_ratio=0` I believe).

All_Relationships/9.entity_prediction.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+# In[9]:
+
+
+get_ipython().run_cell_magic(u'time', u'', u'\n# Train on data without LSTM input\nlstm_features = [\n    "avg_marginal", "quantile_zero", \n    "quantile_twenty","quantile_forty",\n    "quantile_sixty", "quantile_eighty", \n    "lower_ci"\n]\n\n\ntempX = X[[col for col in X.columns if col not in lstm_features]]\ntempX = tempX.append(X_dev[[col for col in X_dev.columns if col not in lstm_features]])\n\nfor models, grids in zip([lr, elastic_net], [lr_grid, elastic_net_grid]):\n    final_model = GridSearchCV(models, grids, cv=10, n_jobs=3)\n    final_model.fit(tempX, Y.append(Y_dev))\n    final_models.append(final_model)')
```
dhimmel: We need to set `scoring='roc_auc'` or [another metric](http://scikit-learn.org/stable/modules/model_evaluation.html#scoring-parameter) that is better than accuracy.

All_Relationships/9.entity_prediction.py
```diff
@@ -0,0 +1,230 @@
...
+# Here we use gridsearch to optimize both models using 10 fold cross validation. After exhausting the list of parameters, the best model is chosen and analyzed in the next chunk. 
+
+# In[8]:
+
+
+n_iter = 100
+final_models = []
+
+lr = LogisticRegression()
+elastic_net = SGDClassifier(penalty='elasticnet', loss='log', random_state=100)
```
dhimmel: Both `lr` and `elastic_net` are logistic regression. They seem redundant. We should get `elastic_net` to perform as good or better than `lr` and then drop `lr`?

dhimmel (commented):
Let's have an in person discussion to go over the important results in detail.

Why no output in the `8.entity_level_features_no_pubmed.ipynb`?

dhimmel (approved):

PR #35, opened 2018-01-18
Python3 upgrade

danich1:
Finally Ready to leave Python 2.7 behind and walk into python 3.6!! This will break something down the road but I'm prepared to fix that.


environment.yml
```diff
@@ -16,19 +13,23 @@ dependencies:
...
+- tqdm=4.19.4
+- conda-forge::tika=1.15
 - pip:
     - backports.lzma==0.0.8
     - git+https://github.com/HazyResearch/numbskull@ac52265038bac8edca3f8e930eff34ebaef4c7a0
+    - hetio==0.2.6
     - lxml==3.6.4
-    - neo4j-driver==1.3.1
-    - tensorflow==1.2.0
+    - neo4j-driver==1.5.3
```
dhimmel: newline

dhimmel (commented):
Yeah stuff will probably break. I think it therefore makes sense to do as many upgrades as possible.

I recommend switching the channel to conda-forge for everything. conda-forge has better availability now than anaconda for most packages and is an open community driven resource.

I would go through each package and find the latest version in conda-forge.

environment.yml
```diff
@@ -1,34 +1,37 @@
...
-- six=1.10.0
-- sqlalchemy=1.1.11
-- spacy=1.8.2
-- tqdm=4.14.0
-- conda-forge::tika=1.14.1
+- numba=0.36.2
+- numpy=1.13.0
+- pandas=0.22.0
+- psycopg2=2.7.3.2
+- python=3.6.2
```
dhimmel: There is more recent. 3.6.4
https://anaconda.org/conda-forge/python

environment.yml
```diff
@@ -1,34 +1,37 @@
...
+- six=1.11.0
+- sqlalchemy=1.1.13
+- sqlite=3.20.1
+- spacy=2.0.4
+- tensorflow==1.4.0
+- tqdm=4.19.5
+- tika=1.15
 - pip:
-    - backports.lzma==0.0.8
     - git+https://github.com/HazyResearch/numbskull@ac52265038bac8edca3f8e930eff34ebaef4c7a0
```
dhimmel: This should be upgraded to latest probably? https://github.com/HazyResearch/numbskull/commit/52192383b30f2835f5b1e17c6365e2f8c93d97da

environment.yml
```diff
@@ -1,34 +1,37 @@
...
-- scikit-image=0.13.0
-- scikit-learn=0.18.1
-- scipy=0.19.0 
-- six=1.10.0
-- sqlalchemy=1.1.11
-- spacy=1.8.2
-- tqdm=4.14.0
-- conda-forge::tika=1.14.1
+- numba=0.36.2
+- numpy=1.13.0
```
dhimmel, line 18: https://anaconda.org/conda-forge/numpy
1.14

dhimmel (commented):
Some of the packages still aren't the latest. Feel free to merge without additional review after the comments are addressed.

environment.yml
```diff
@@ -1,34 +1,37 @@
...
+- py4j=0.10.6
+- requests=2.18.4
+- seaborn=0.8.1
+- scikit-image=0.13.1
+- scikit-learn=0.19.1
+- scipy=1.0.0 
+- six=1.11.0
+- sqlalchemy=1.1.13
+- sqlite=3.20.1
+- spacy=2.0.4
```
dhimmel: The snorkel [`python-package-requirement.txt`](https://github.com/HazyResearch/snorkel/blob/b2655a4e6de79a945d08de004297b130585c76d5/python-package-requirement.txt#L19) pins `spacy=1.10.0` so we should probably stick with that.

dhimmel (changes requested):
Also make sure the environment installs.

environment.yml
danich1: I get this error when using numpy 1.14.0:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - numba 0.36.2* -> numpy >=1.13,<1.14.0a0
  - numpy 1.14.0*
Use "conda info <package>" to see the dependencies for each package.
```
As a result I stuck with numpy 1.13.0 because that installed fine.

environment.yml
danich1: **Just a side note**: conda-forge doesn't have this specific version of spacy and neither does the default, so it is filed under the pip install section.

dhimmel (approved):

PR #36, opened 2018-02-05
Updated Entity Level Prediction Module 

danich1:
This PR is a mix of bug fixing and adding data files onto the repo. I fixed the naming convention on the data file, so hopefully they make more sense. Furthermore, I calculated AUROCs for each feature which is located in notebook 9. Let me know what you think.


All_Relationships/9.entity_prediction.py
```diff
@@ -151,25 +154,40 @@
...
 
 
+feature_rocs = []
+for feature in X.columns:
+    fpr, tpr, _ = roc_curve(Y, X[feature])
+    feature_auc = auc(fpr, tpr)
+    feature_rocs.append((feature, feature_auc))
+
+feature_roc_df = pd.DataFrame(feature_rocs, columns=["Feature", "AUROC"])
+sns.barplot(x="AUROC", y="Feature", data=feature_roc_df)
```
dhimmel: Would be nice to change x-axis to go from 0.5 to 1

dhimmel (commented):
It troubles me that the purported high performance of the LSTM-based features is not preserved in the models. Let's dive into this in person on Tuesday

dhimmel:
Let's update the model to use only a single LSTM feature. It looks like the mean or 80th percentile would be the best choice. Let's see if this helps fix the issue where the model is worse than a single feature.

Regarding the correlation plot, usually you'd want to use a diverging colormap because correlations go from -1 to 1. However, since there aren't many negative correlations here, viridis is fine.

dhimmel:
We should be standardizing the features using an sklearn pipeline. Cognoma notebook has [an example](https://github.com/cognoma/machine-learning/blob/master/2.mutation-classifier.ipynb), although it's actually its more complex than what we need.

Additional references:

+ https://stats.stackexchange.com/a/290960/74908
+ http://scikit-learn.org/stable/auto_examples/preprocessing/plot_scaling_importance.html

All_Relationships/9.entity_prediction.py
```diff
@@ -148,28 +167,62 @@
...
 
 
-zip(final_models[1].best_estimator_.coef_[0], [col for col in training_set.columns if col not in non_features])
+# # Corerlation Matrix
+
+# In[18]:
+
+
+feature_corr_mat = train_X.corr()
+sns.heatmap(feature_corr_mat, cmap="RdBu")
```
dhimmel: Must add `center=0` so colors diverge from 0!!!

dhimmel (approved):
This is a very important PR and sets the stage for us to be able to evaluate the contribution of LSTM NLP features going forward. My comments below are in reference to [`9.entity_prediction.ipynb`](https://github.com/danich1/snorkeling/blob/696591cb2eac016f6dbe17471f1fc21f40db26a7/All_Relationships/9.entity_prediction.ipynb)

Some key findings here are the AUROCs (train / test) I notice:

+ Node degree prior = 71% / 72%
+ Term co-occurrence p-value = 75% / 75%
+ mean LSTM NLP probability = 94% / 71%

Now for multi-feature regression models:

+ prior + co-occurrence = 87% / 87%
+ prior + co-occurrence + LSTM NLP = 96% / 82%

My interpretation. Adding co-occurrence to the prior improves testing AUROC from 75% to 87%. However, adding LSTM decreases it to 82%. This is because the LSTM has overfit on the training data. Accordingly, It's being assigned too high of a weight in the combined model thereby increasing training performance to 96% from 94% but testing performance suffers.

Tagging @cgreene. These results are encouraging because they give us the diagnostics we need, but we see that we need to prevent the LSTM from overfitting. We will have to interrogate how we train the LSTM and on what observations to know exactly why it's overfitting.

Note for future, we will want to transform the prior probability with a logit function when using it in a model. Example [here](https://github.com/dhimmel/learn/blob/7668c97b2a6f348479b70fa40c3d7db424584315/prediction/4-predictr.ipynb), also noting [this comment](https://think-lab.github.io/d/194/#3).

@danich1 go ahead and squash merge when you're ready.

dhimmel:
https://github.com/greenelab/snorkeling/pull/36/commits/40ca5bd55660f068ab97408b8858137d94d9919d logit not log. Time for a [scipy special](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.logit.html). Logit enables logistic regression to regurgitate the prior probability _ad infinitum_

PR #38, opened 2018-03-25
Adding updated lstm results

danich1:
Quick pr for adding updated LSTM results. Ran for 22 epochs and slowly see an increase in test predictions. The other files edited are small bug fixes I encountered while generating the LSTM results.


dhimmel (approved):

dhimmel:
Do we keep track of training and testing performance (sentence-level) after each epoch? It would be nice to visualize how much each epoch changes performance.

danich1:
Their script outputs the loss for each epoch. I don't have the values saved for this run, because of gpu cluster issues. Following the downstream experiments, I'll be saving that information as well.

dhimmel:
Great. @danich1 will let you merge this when you're ready.

PR #41, opened 2018-05-01
Added new stratification strategy

danich1:
This PR contains code for our new stratified sampling strategy.  This approach calculates a ranking for each disease-gene pair grouped by its presence in hetionet and pubmed. From this ranking one can extract certain pairs based on the desired split size for training, dev and testing sets. The bottleneck here is updating the rows of the candidate table. Takes ~3 hrs.


dhimmel (approved):
Great. Nice to see the database update method worked. The table update still takes a long time (3 hours). If this becomes a problem in the future we could look into dropping all affected rows and uploading them fresh.

PR #42, opened 2018-05-01
Snorkel Checkpoint - Modeling Label Functions

danich1:
This PR is a checkpoint for the first day of @dhimmel and I's snorkel adventure. In congruence with our new modeling strategy we changed some fo the label functions with a 0 rather than a 1. Furthermore we decided to split our hetionet knowledge base label function (LF_HETNET) into individual source specific label functions. This helped with performance, but more to come.


dhimmel (approved):
As stated, in progress, but we want to track our daily progress.

PR #47, opened 2018-05-02
Refactored code to efficiently run LSTM on cluster

danich1:
This block of code here is designed to run the LSTM on the cluster. We need to do this in order to have access to the GPU UPenn has. The script files are written in python 2.7.12. If we want to upgrade to python 3, it might take a bit of time. Plus, this should be moving to pytorch shortly.


dhimmel (approved):
Python 3 would be nice, but we can wait till the pytorch migration.

PR #49, opened 2018-06-05
Gen model calibration

danich1:
Added new label functions for the generative model. The results of this model will be shown in the next pull request. Let me know what you think.


All_Relationships/utils/disease_gene_lf.py
```diff
@@ -210,35 +235,64 @@ def LF_ASSOCIATION(c):
...
+    ]) else 0
+
+def LF_DG_DISTANCE_SHORT(c):
     """
     This LF is designed to make sure that the disease mention
     and the gene mention aren't right next to each other.
     """
-    token_in_btw_len = len(get_text_between(c).split(" "))
-    return -1 if token_in_btw_len < 5 else 0
+    return -1 if len(get_text_between(c).split(" ")) < 5 else 0
```
dhimmel: Why not <=2... for example "gene was associated with disease" is 3.

All_Relationships/2.candidate-labeler.py
```diff
@@ -44,24 +44,25 @@
 session = SnorkelSession()
 
 
-# In[3]:
+# In[ ]:
 
 
 from snorkel.annotations import FeatureAnnotator, LabelAnnotator
 from snorkel.features import get_span_feats
+from snorkel.lf_helpers import *
```
dhimmel: Is there an extraordinary circumstance that justifies using `import *`?

All_Relationships/utils/disease_gene_lf.py
```diff
@@ -132,25 +133,35 @@ def LF_CHECK_DISEASE_TAG(c):
     """
     sen = c[0].get_parent()
     disease_name = re.sub("\) ?", "", c[0].get_span())
+    disease_name = re.sub(r"(\w)-(\w)", r"\g<1> \g<2>", disease_name)
```
dhimmel, line 136: If speed is an issue, i.e. the regexes are the slowest part, you may consider compiling the regexes.

dhimmel (approved):
Just some minor comments.

These labeling functions are great for unit tests. I recommend pytest. With the tests you'd pass some example sentences to make sure the LFs label them as you'd anticipate.

Writing tests takes time and I'm not 100% saying the time would be worth the benefits here, so @danich1 up to you whether that's something you're interested in

All_Relationships/2.candidate-labeler.py
danich1: Opps. I left that in there when I was debugging one of the label functions. Removing now...

All_Relationships/utils/disease_gene_lf.py
danich1: So far I haven't reached a bottleneck with this yet; however, will keep in mind next round of working on the label functions.

PR #50, opened 2018-06-06
Gen model calibration

danich1:
In this pr I have the results of running two generative models on a 50k training set. The first model assumes each label function is conditionally independent and the second model takes correlation structure into account (dependency aware). It turns out that modeling the correlation structure is better for performance than assuming independence. Let me know what you think.


dhimmel (commented):
As @danich1 and I discussed, this is a nice comparison between DA and CI (conditionally independent). It would be best to compare using something that assesses multiple binarization thresholds such as AUROC.

dhimmel (approved):
The ROC curve is helpful... neither labeling method is producing accurate labels. They are both just barely better than random. I think much of this is the difficulty in labeling sentences of disease-gene associations. Even labeling these sentences by manual curation is incredibly difficult.

PR #53, opened 2018-06-27
Co-occurence analysis with dataset

danich1:
This PR contains the results from the co-occurrence analysis. Notebook 9 contains the ROCs and the needed graphs for this analysis. From the looks of things adding the NLP marginals does not make a big difference in performance. Let me know what you think.


danich1:
Due to realization of an index error in previous notebooks this PR is now incorrect. Going to close and reopen the correct version sometime this week or next.

PR #54, opened 2018-07-02
Snorkel upgrade to 0.7.0 beta

danich1:
This PR upgrades snorkel to 0.7.0. In regards to this change, there is no longer a need to set environmental variables via bash commands. Also, pytorch is now available, which means we can have better customization in terms of using recurrent neural networks.


dhimmel (commented):
Looks good and like it was pretty easy of an upgrade. Have you tested that the new environment actually works? It's possible it may take a little while to iron out any lingering issues.

I think we should be versioning the pip installs with commit hashes. This is essential to having a reproducible environment, especially since those packages are likely to have breaking changes.

danich1:
I have tested the new environment and it seems to be working smoothly.

dhimmel (approved):

PR #55, opened 2018-07-02
Gen model evaluation - Mismatch fix

danich1:
This pull request consists of updated results, after accounting for an index mismatch bug I found in the code. Candidates are extracted from the database in sorted order (by their ids); however, when I created an excel file to look at the development set's candidates, I shuffled the order. Since I didn't account for this, the ROCs and the PRs showed poor performance. Now that I have remedied this issue, some of the results are looking a bit better. Will add the discriminator performance in the next pull request.


dhimmel (approved):
Performance does look better

> after accounting for an index mismatch bug I found in the code

If possible, I always to to avoid any type of position-dependent indexing because this type of error is so common and difficult to detect. Not sure whether alternatives are available in this case, but consider writing code that CANNOT develop this sort of bug.

PR #56, opened 2018-07-02
Updating Discriminator Model Performance

danich1:
This PR is the second part of the whole mismatch debacle (#55). With this fix we can see that Doc2Vec using all of Pubtator does the best in terms of prediction performance here. More work needed, but adding updated results here.


cgreene:
Just noting that the notebooks all still say LSTM even though I don't think you're using an LSTM anymore. Probably good to update that.

dhimmel (approved):

PR #57, opened 2018-07-26
Improved Generative Model Performance

danich1:
This PR contains updated label functions and shows an increase in the performance of the generative model.


dhimmel (approved):
The generative model does look to have improved considerably. Hopefully, this will yield improved discriminator models as well.

Is this still on Disease-Gene associations?

Would be helpful for @danich1 to go over what has been modified to improve performance with me when I'm back in town, but might as well go ahead and merge now.

danich1:
Yes this is on Disease-Gene associations. We can talk more when you get back.

PR #58, opened 2018-08-01
Added results for 17 v 26 lfs

danich1:
This PR is the second half where it shows the discriminator model performance using 17 lfs and 26 lfs.


dhimmel (approved):
Additional labeling functions appear to help.

PR #59, opened 2018-08-17
Changed Notebook and Re-organized Repository for Sanity Purposes

danich1:
This pull requests adds data files that I neglected to add in from previous PRs and rearranges a few files for sanity purposes. I also made small changes to the 2nd notebook for the next PR.


dhimmel (approved):
Housekeeping always a good idea!

One pointer if you used pandas to save `gen_model_marginals_history.csv`, in future you may like `float_format='%.5g'` to do nice rounding. In many cases we don't care about the numbers to max precision and it just harms readability.

PR #60, opened 2018-08-17
Russ Data Experiment

danich1:
This PR contains data from the paper "A global network of biomedical relationships derived from text" [paper here](https://zenodo.org/record/1035500#.W3Hc-RgpBrk)). We used these clusters themes as features to aide in our goal of predicting disease associates with gene (DaG) relationships. Unfortunately these clusters themes do not seem to have predictive power in our case. 

**TLDR Paper Summary Here**: This paper focused on using the bi-clustering algorithm on a matrix dependency paths that were generated from every sentence in Pubtator (version Sept 15-2017). After these clusters were generated, manual curation was involved to assign "themes" to each generated cluster. Lastly, theme distributions were calculated and certain paths were manually identified to be central to their overall theme.

**Themes used for this experiment:**

| Theme Abbreviation | Theme |
| ------------- | ------------- |
| U | Causal Mutations | 
| Ud | mutations affecting disease course |
| D | drug targets |
| J | role in pathogenesis |
| Y | polymorphisms alter risk |
| G | promotes progression | 
| Md | biomarkers (diagnostic)|
| X | overexpression in disease |
| L | improper regulation linked to disease|

### DIFF
--- a/All_Relationships/9a.entity_prediction_russ_experiment.py
+++ b/All_Relationships/9a.entity_prediction_russ_experiment.py
@@ -1,0 +1,518 @@
+
+# coding: utf-8
+
+# # Train/Test the True Relationship Model
+
+# This notebook is design to predict DG relationships on the entity level. Here we are taking the input from this [paper](https://zenodo.org/record/1035500#.W3Hc-RgpBrk), prior probability notebook and the summary statistics notebook, then combinding it into a single dataset. From there we train a Ridge LR model and an elastic net LR model to make the final prediction.
+
+# In[1]:
+
+
+get_ipython().run_line_magic('load_ext', 'autoreload')
+get_ipython().run_line_magic('autoreload', '2')
+get_ipython().run_line_magic('matplotlib', 'inline')
+
+from collections import defaultdict
+import csv
+import os
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import tqdm
+import re
+from scipy.stats import fisher_exact
+from scipy.special import logit
+import scipy
+from sqlalchemy import and_
+from sklearn.linear_model import LogisticRegression, SGDClassifier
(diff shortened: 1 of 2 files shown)


cgreene:
Just to check - this doesn't look like it includes trying them as labeling functions. Is that correct?

danich1:
Yes this is just using them as features. Using them as label functions will come later.

All_Relationships/9a.entity_prediction_russ_experiment.py
```diff
@@ -0,0 +1,518 @@
...
+    .merge(theme_dist_df.astype({"path":str}), left_on='dependency_path', right_on='path', how='left')
+)
+final_dg_df.head(2)
+
+
+# In[9]:
+
+
+# Transform the nans into 0.0 and aggregate the other sentnece scores
+final_dg_df[['U', 'U.ind','Ud', 'Ud.ind',
```
dhimmel, line 134: What does `.ind` mean?

Are these two lists of columns the same? If so, I'd just define them once as a variable above and use that twice. DRY.

dhimmel (commented):
Nice plots regarding the feature correlations, feature informativeness, and model performance.

Question regarding the `.ind` features.

I am OKAY not reviewing this in great depth, if we don't think it'll be an essential part of the snorkeling pipeline. So let me know if you want a more in depth review, otherwise merge at will.

All_Relationships/9a.entity_prediction_russ_experiment.py
danich1: The .ind column is a binary variable that indicates if a dependency path is regarded as highly significant in the appropriate category. The naming was suppose to be intuitive, but I can change it if desired.

danich1:
No need for an in depth review in this case. This was a small experiment to see how informative Russ's data on Pubtator was in terms of predicting disease associates with gene relationships. More important experiment will be coming this week.

dhimmel (approved):

PR #61, opened 2018-08-20
Added results for 250k exp

danich1:
This PR follows #59, where we tested to see if performance improves with the increase in amount of data provided. After some run-time, we reached the conclusion that performance depends on the label functions rather than the amount of data.


dhimmel (approved):
The learned accuracy plot caught my eye. I like how the dependency aware model favors certain LFs to a higher degree than the conditionally independent model. However, the PR and ROC curves don't indicate this affected performance in a meaningful way.

PR #62, opened 2018-08-23
CbG Experiment + Bicluster Data LFs

danich1:
This PR tested the hypothesis that DaG label functions will transfer onto CbG sentences.  Notebook 3 has results for this experiment. As one can see DaG lf doesn't add towards the performance. The next experiment will be seeing how well both DaG and CbG functions work on Chemical Treats Disease relationship.


dhimmel (commented):
I don't love that the DaG notebooks are being overwritten. Ideally we would have separate notebooks or cells or loops to accommodate different edge types. If you find code is being repeated, then perhaps that code should be put into modules.

I'm not going to say that this PR shouldn't proceed overwriting DaG work, but that you should think about modularity and supporting multiple relationship types simultaneously going forward. This can be difficult with noteobooks, so you may have to modularize some code into .py files. 

> The next experiment will be seeing how well both DaG and CbG functions work on Chemical Treats Disease relationship.

Regarding CbG labeling functions, I'd recommend one per source as we do for DaG Hetionet-derived LFs.

This will be my last review before my travels. So while I'm gone I don't think it makes sense to wait on my approval for PRs. Proceed as you see fit.

danich1:
Great Point. I made changes to reflect this suggestion. Since @dhimmel is out of town, I am adding @ajlee21 to review this PR.

All_Relationships/2.candidate-labeler.py
```diff
@@ -146,7 +140,27 @@
...
+labeler = LabelAnnotator(lfs=list(CG_LFS["CG_DB"].values()) + 
+                         list(CG_LFS["CG_TEXT"].values()) +  
+                         list(CG_LFS["CG_BICLUSTER"]) + 
+                         list(DG_LFS["DG_TEXT"].values()))
+
+
+# In[ ]:
+
+
+def make_sentence_df(candidates):
```
ajlee21, line 151: I would suggest adding comments for your different functions so that when you return to these scripts later you remember what they were input/output is and what they were used for

ajlee21: Same for get_split that is above

All_Relationships/utils/datafiles/compound_gene_datafiles.py
```diff
@@ -0,0 +1,64 @@
...
+)
+
+gene_entrez_df['dummy_key'] = 0
+compound_df['dummy_key'] = 0
+pair_df = (
+	gene_entrez_df
+	.merge(compound_df[["drugbank_id", "name", "dummy_key"]], on='dummy_key')
+	.drop('dummy_key', axis=1)
+)
+pair_df.to_csv("data/compound_gene/compound_binds_gene/compound_gene_pairs_binds_full_map.csv", index=False, float_format='%.5g')
```
ajlee21: Minor cleanup - I would create a filename variable that you can easily change in the beginning of the code.  Not sure if this file will always be stored/named the same because in that case perhaps you don't want the user to accidentally have access and then change it

All_Relationships/utils/disease_gene_lf.py
```diff
@@ -89,7 +89,7 @@ def LF_HETNET_DisGeNET(c):
 def LF_HETNET_GWAS(c):
```
ajlee21, line 101: I would try to be more consistent about having comments for what each of the label functions are doing and also what the category is too

All_Relationships/3.model-label-functions.py
```diff
@@ -149,59 +148,105 @@
 # In[12]:
 
 
-get_ipython().run_cell_magic('time', '', '#Conditionally independent Generative Model\nindep_gen_model = GenerativeModel()\nindep_gen_model.train(\n    L_train,\n    epochs=10,\n    decay=0.95,\n    step_size=0.1 / L_train.shape[0],\n    reg_param=1e-6,\n    threads=50,\n)')
+def get_columns(lf_hash, lf_name):
```
ajlee21, line 151: same comment as above

All_Relationships/3.model-label-functions.py
```diff
@@ -149,59 +148,105 @@
...
 
 
 # In[17]:
 
 
-plt.hist(train_marginals_indep, bins=20)
-plt.title("CI Training Set Marginals")
-plt.ylabel("Frequency")
-plt.xlabel("Probability of Positive Class")
+f, axs = plt.subplots(len(lfs_columns), 1, figsize=(10, 5), sharex=True)
```
ajlee21: I might not understand the meaning of the plots here when we went over them together so I don't know if you want to add comments here/change the title of the axes

All_Relationships/3.model-label-functions.py
```diff
@@ -236,195 +289,266 @@
...
-plt.title("ROC Curve of Generative Models")
-plt.xlabel("False Positive Rate")
-plt.ylabel("True Positive Rate")
-plt.legend()
+#gen_model_history_df['lfs_26'] = L_dev_ci_marginals
+#gen_model_history_df.head(2)
 
 
-# In[ ]:
+# In[34]:
```
ajlee21, line 421: Why do you have this block commented out?

ajlee21 (approved):
Basically just a bunch of minor edits: adding comments to give clarity

All_Relationships/3.model-label-functions.py
danich1: That is blocked out because I didn't want to run that section of the notebook for this experiment. Granted I'm sure there is a better way to handle this situation, but it sufficed for now.

cgreene: This is lightly worrisome and, like @dhimmel, makes me wonder if some of the code in these notebooks should really be functions that are imported and called.

danich1: Ah ok. Don't want to blow up this PR with a lot of changes, so I am going to upload the changes I made (per the above comments) and then open a new PR for some code restructure and housekeeping.

PR #63, opened 2018-08-31
Housekeeping

danich1:
This PR is designed to perform some housekeeping items. Majority of this involves file moving and renaming. To save some time and headache please only consider the scripts with the .py extension for review. The notebooks are blank on purpose, because they will be filled in the next PR after I get results from the compound_treats_disease experiments.


All_Relationships/2.candidate-labeler.py
```diff
@@ -47,246 +42,170 @@
...
+train_df = pd.read_excel(spreadsheet_names['train'])
+train_df.head(2)
 
 
 # In[ ]:
 
 
-train_df = make_sentence_df(candidates)
-train_df.head(2)
+train_candidate_ids = list(map(int, train_candidate_df.candidate_id.values))[0:10]
```
ajlee21: Looks like you have some hard coded values here.  Will this stay fixed in the future?

All_Relationships/2.candidate-labeler.py
```diff
@@ -16,16 +16,11 @@
...
 import os
-import re
 
-
-import matplotlib.pyplot as plt
-import numpy as np
 import pandas as pd
-from tqdm import tqdm_notebook
+
+from utils.notebook_utils.dataframe_helper import make_sentence_df, write_candidates_to_excel
```
ajlee21: Add description at the header for what this script is doing/contains

All_Relationships/3.model-label-functions.py
```diff
@@ -9,25 +9,22 @@
...
 
-import matplotlib.pyplot as plt
-import numpy as np
 import pandas as pd
 import seaborn as sns
 from sklearn.metrics import average_precision_score, precision_recall_curve, roc_curve, auc, f1_score
 
 
-# In[2]:
+# In[ ]:
```
ajlee21, line 28: description here

All_Relationships/3.model-label-functions.py
```diff
@@ -145,37 +146,27 @@
...
-dg_text = get_columns(DG_LFS, "DaG_TEXT")
+lfs_columns = [
+    cg_text
+]
 
-lfs_columns = [cg_db, cg_text,dg_text, cg_db+cg_text, cg_db+cg_text+dg_text]
-model_names = ["CbG_DB", "CbG_TEXT", "DaG_TEXT", "CbG_ALL", "DaG_CbG_ALL"]
+model_names = [
+    "CbG_TEXT"
+]
```
ajlee21: Will you be running multiple models at one time (i.e. will the list every have a length > 1)?  Maybe add a comment with the list of possible model choices

All_Relationships/utils/notebook_utils/bicluster.py
```diff
@@ -0,0 +1,44 @@
+import pandas as pd
+import re
+
```
ajlee21, line 3: This my be a stylistic preference, just curious if there is a reason why you have divided the python functions into different files as opposed to having all the definitions in a single file with comments for the different functionalities.

ajlee21 (approved):
Mostly minor comments.  Nice improvements from what I can see without knowing too much about the global structure of this repo

All_Relationships/2.candidate-labeler.py
danich1: Ack! Good catch that was old code that needed to be changed

All_Relationships/utils/notebook_utils/bicluster.py
danich1: Main reason for this is that my mind works better when functions are in specified files compared to just a single universal file. Also, it helps with my organization of this repository.

PR #66, opened 2018-09-24
Generative Model Benchmarking and bug fixes.

danich1:
This is PR is part of the refactoring process. There are minor script changes to fix bugs or add needed functions. The bigger change is the notebook that is being uploaded. It contains the results for  bench-marking the generative model. Finally figured out the problem, which was that there wasn't enough regularization and the model weights were doing whatever they wanted. Now that has been solved, I have achieved aurocs close to 70%.


disease_gene/disease_associates_gene/experiments/gen_model_benchmarking.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+from snorkel.annotations import LabelAnnotator
+from snorkel.learning.structure import DependencySelector
+from snorkel.models import candidate_subclass
+
+from utils.label_functions import DG_LFS
+
+from utils.notebook_utils.dataframe_helper import load_candidate_dataframes
+from utils.notebook_utils.label_matrix_helper import *
```
dhimmel: Let's avoid `import *` as it can result in ambiguity

disease_gene/disease_associates_gene/experiments/gen_model_benchmarking.py
```diff
@@ -0,0 +1,329 @@
...
+
+get_ipython().run_line_magic('load_ext', 'autoreload')
+get_ipython().run_line_magic('autoreload', '2')
+get_ipython().run_line_magic('matplotlib', 'inline')
+
+from itertools import product
+import os
+import sys
+
+sys.path.append(os.path.abspath('../../../.'))
```
dhimmel: Can we put `utils` in a directory with all your module code, such that python isn't searching every subdirectory of the root repo for code? For example, `modules/utils`.

dhimmel: `../../../modules`

dhimmel (commented):
Nice to see that the generative model to produce consensus labels is improving with better regularization.

Consider whether we need to spend CV-time optimizing burn-in and epochs? Or whether instead we can pick a value that is high enough and use the CV savings to evaluate more regularization options.

Also let's keep an eye out for instances where the same parameter combinations lead to drastically different performance estimates, and let's try to change burn-in / epochs to avoid these "bad rolls of the die".

Would be nice to plot CV-AUROC in a way that shows regularization as a visual parameter... ie scatterplot or heatmap.

dhimmel:
In the following plot, is the x-axis the regularization parameter?

auc

At low values, the performance is erratic. Erratic is bad. Why is the performance erratic? Personally 0.4 looks best to me, but I'd like to understand how the erratic performance is sometimes giving such high AUROC.

Let's chat.

dhimmel (approved):
Nice regularization optimization plots! Seeing these plots makes me feel more confident that we can pick a near-optimal and stable choice for this parameter.

PR #67, opened 2018-09-24
added degree-prior.csv

danich1:
I forgot to add a file that was generated a while back for generating prior edge probabilities. So adding now..


dhimmel (approved):
Nice! This will be helpful for our investigation of an analytic method for determining XSwap prior probabilities as per https://github.com/greenelab/hetmech/issues/134#issuecomment-423765409.

PR #68, opened 2018-10-03
Discriminator Model Results for DaG sentences

danich1:
This PR contains results for the various discriminator models. Only major file to review here is the notebook and the results. Let me know what you think.


dhimmel (approved):
Nice pull request. @danich1 and I discussed the results in person.

Two thoughts on my end:

1. let's double check that `run_grid_search` is using auroc rather than unbalanced accuracy as the performance metric.
2. Currenty the dev set only havs ~60 hand labeled sentences. This is too few, and could result in volatile AUROC estimates. It would be good to rerun the main notebook here with a bigger dev set, so performance could be assessed more accurately.

PR #70, opened 2018-11-06
Metal Integration and Updated Generative Model

danich1:
I realized that I haven't updated this repo, which is now being remedied. This pull requests adds snorkel metal integration, which will be used for the machine learning models from now on. I fixed some bugs and typos in my code. Plus, I added a new test set which makes more sense moving forward. For this pull request, the main file that needs to be reviewed is **gen_model_benchmarking.ipynb**.


dhimmel (approved):
parameter tuning is looking better than before (more stable)

PR #75, opened 2019-02-13
Single Task Gen Model Update

danich1:
This pull request is designed to update the repository with the new findings for the generative model. Only concern yourself with the notebook as the other scripts and files are there to help run cells in the notebook. Let me know what you think.


dhimmel (commented):
Went over in person with David today. A few TODOs and then it looks good to me.

dhimmel (approved):

PR #78, opened 2019-04-16
Using Label Functions Designed for One Relation to Predict Another Relation

danich1:
This pull request is designed to upload the results of my latest experiment, where I aimed to see if one relation type can aid in predicting another relation type. I.E Can compound treats disease label functions classify sentences that infer Disease associating a Gene relationship? How about compound binds gene label functions predicting sentences that infer disease associates a gene relationship? etc...

To answer these questions the approach here was to randomly add label function information to the baseline model (a distantly supervised classifier using relation specific database to label sentences) and estimate if performance improved or worsened. 

There are a lot of files being added here; however, the only files that need to be reviews are the `single_task_label_function_experiment.ipynb` notebooks. There is a notebook for each relation type. Let me know what you think. If you get a chance take a look at the images within the figures folder. It will provide a global picture on what I had accomplished.


ajlee21 (approved):
1.  Would change the title from "compound disease" to "compound treats disease"
2.  Instead of saying "GiG vs CtD" I would rephrase it to say "Using GiG labels to predict CtD claims" so its more clear.
3.  Would be nice to have a brief description on what the 4 different plots are showing

Interesting results based on the summary figures.  I remember we discussed this a little bit.  Its interesting that the label functions appear to help make predictions in other categories.  However the DaG category appears be be worse with more labels (AUROC)

Looks like you get fairly consistent results with PR and AUC except for CtD.

cgreene:
The title is transfer learning, but I'm not sure if that's quite the right term. Maybe `label transfer`?

It's really pretty amazing how robust things are in AUC terms to moving labels around. It's also interesting how much having some text-based LF helps, somewhat (apparently) regardless of what exact LF it is. Cool!

PR #79, opened 2019-05-10
Class Accuracy for Label Function Random Sampling Experiment

danich1:
This pull request is a follow up to the [previous pr](https://github.com/greenelab/snorkeling/pull/78), where I aimed to see if label functions designed for one relation can predict another relation. Inside this pr, I plotted individual class accuracy (how well do these label functions predict the positive and negative classes). Main figures to observe: class_correct_dev_set.png and class_correct_test_set.png. 

Feel free to take a look at the code if desired, but it all should relatively familiar in reference towards teh last pr.
@ajlee21 I know you have prelims coming up. No worries if you don't have time.


ajlee21 (approved):
Very nice PR.  Just a few questions to clarify mu understanding:

1.  What does your table with the polarity, coverage, overlap showing?
2.  For your AUROC and AUPR.  This is showing that your ability to predict say GiG using non-GiG labels increases as you add label function?
3.  Think we discussed this before that it might help to interpret results if the "0" = baseline generative model only is displayed as a horizontal threshold to beat
4.  So for your % correct of positive and negative associations, it looks like the number of label functions affects the % you correctly label as positive not negative as much.  I guess that's not too surprising given that your label functions are meant to identify positive associations.  Would've kinda expect the curves to act inversely though (i.e. as one goes up the other goes down).

danich1:
As discussed in person:
1. The table here describes the label function matrix in terms of directionality (positive or negative: polarity), amount of sentences a label function covers (coverage), how often does one label function appear with other label functions (overlap) and how often does one label function output labels that conflict with another label function (conflict)
2. In the curves the right hand side are the relations being predicted and the column titles are the different types of sources being used. 
3. I originally had that depiction; however, I believe the feed back was for me to have the baseline as a point rather than a line. Sticking with the current graph until further notice.
4. Most of the curves have an inverse relationship. Just some of graphs have a small difference compared to others.

PR #82, opened 2019-06-02
Added Results and Updated Figure Generating Code

danich1:
This PR is adding all the data files used in the label sampling experiment. I also updated the figure generating code based on lab feedback and updated a line in one of my modules to make code work in my notebook.

Should be a very quick review. Only have to look at the py files for the figures. Let me know what you think.


figures/label_sampling_experiment/plot_scripts/class_accuracy_figure_generator.py
```diff
@@ -0,0 +1,123 @@
...
+
+plt.switch_backend('agg')
+
+def get_dataframes(result_dir, file_path):
+    return {
+        # Get the head word of each file that will be parsed
+        os.path.splitext(os.path.basename(file))[0].split("_")[0]:
+        pd.read_csv(file, sep="\t")
+        .assign(num_lfs=lambda x: x['num_lfs'].map(lambda y: y if y != 'baseline' else 'BL'))
+        .query("num_lfs in [0,1,6,11,16,20,28,22,30] or num_lfs in [0,1,33,65,97,100]")
```
ajlee21: Why are there two different ranges here?

figures/label_sampling_experiment/plot_scripts/class_accuracy_figure_generator.py
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+
+def plot_performance_graph(metric='AUROC', evaluation_set='dev', title="", file_name="", data=None):
+    fig, axes = plt.subplots(len(file_tree), len(file_tree["DaG"]), figsize=(25, 15), sharey='row')
+
+    for row_ind, col in enumerate(data):
+        
+        for col_ind, row in enumerate(data[col]):
+            sns.pointplot(x="num_lfs", y=metric, data=data[col][row][evaluation_set], ax=axes[row_ind][col_ind])
```
ajlee21, line 68: What is the structure of your input dataset?

figures/label_sampling_experiment/plot_scripts/class_accuracy_figure_generator.py
```diff
@@ -0,0 +1,123 @@
...
+})
+
+
+plot_performance_graph(
+    metric="frac_correct", evaluation_set='dev',
+    title="Individual Class Performance (Dev Set)",
+    file_name="class_correct_dev_set.png", data=class_correct_data_tree
+)
+
+plot_performance_graph(
```
ajlee21: So these .py files are not being called from your main nb?  In terms of organization i usually like to do that.  But I'm not seeing the overall big picture of this repo

figures/label_sampling_experiment/plot_scripts/label_sampling_disc_figure_generator.py
```diff
@@ -0,0 +1,187 @@
...
+    fig.text(0.5, 0.89, 'Label Sources', ha='center', fontsize=30)
+    fig.text(0.5, 0.04, 'Number of Additional Label Functions', ha='center', fontsize=30)
+    fig.text(0.04, 0.5, f'Predicted Relations ({metric})', va='center', rotation='vertical', fontsize=25)
+    fig.suptitle(title, fontsize=30)
+    fig.text(0.69, 0.02, '0-Only Uses Relation Specific Databases.', fontsize=27)
+    plt.subplots_adjust(top=0.85)
+    plt.savefig(file_name, format='png')
+
+
+color_map = {
```
ajlee21: Not sure if you want to have a map using the english word for these colors for future reference

ajlee21 (approved):
Very nice updated figures!  Really think the colors help to clarify your message.

Just a few question on your added figure_generator scripts.  

Also just a note, you may already be doing this, but in preparation for your committee meeting, might want to think about maybe adding a few more comments to break up your nb to explain: what quick-start does, what the difference between dev and test are, etc

figures/label_sampling_experiment/plot_scripts/class_accuracy_figure_generator.py
cgreene: Alex's question is good. Also the numbers are not ascending, not that evenly spaced, and more of an issue is that there's a lot of magic numbers here. Is there a way to do this that doesn't have these properties? I'm imagining 0 is a special case because it's the knowledge bases, but can the rest be generated and selected programmatically?

danich1: Ack magic numbers.. Good point. Updated the code to give the reader more information about these numbers.

figures/label_sampling_experiment/plot_scripts/class_accuracy_figure_generator.py
danich1: The data variable is a dictionary with relations as keys and a dataframe with the following values:

| label | Num_lfs | frac_correct |
| --- | --- | --- |
| 0 | 6 | 0.5808|
| 1 | 6 | 0.4192|

For the generator/discriminator models the data values are:
 
| AUPRC | AUROC | Num_lfs |
| --- | --- | --- |
| 0.419 | 0.568 | 0 |
| 0.418 | 0.56889 | 1|

figures/label_sampling_experiment/plot_scripts/class_accuracy_figure_generator.py
danich1: I like the plan to make these scripts actual notebooks instead. Given the current time constraint, I'll push this idea onto the immediate todo list pile.

PR #84, opened 2019-06-14
Add word vector scripts to repository

danich1:
This PR is designed to upload scripts/notebooks that aid in training word vectors for the discriminator model. The tool we are using is Facebook's Fasttext and generating 300 dimension word vectors.


compound_disease/compound_treats_disease/word_vector_experiment/generate_word_vectors.py
```diff
@@ -0,0 +1,244 @@
+
+# coding: utf-8
+
+# # Generate Word Vectors For Compound Treats Disease Sentences
+
+# This notebook is designed to generate word vectors for gene interacts gene (CtD) sentences. Using facebooks's fasttext, we trained word vectors using all sentences that contain a disease and gene mention. The model was trained using the following specifications:
```
ajlee21: Typo: gene-interacts-gene instead of compound-treats-disease, also disease and compound mention

compound_disease/compound_treats_disease/word_vector_experiment/generate_word_vectors.py
```diff
@@ -0,0 +1,244 @@
...
+# In[9]:
+
+
+cutoff = 300
+total_candidates_df = (
+    pd
+    .read_table("../dataset_statistics/results/all_ctd_map.tsv.xz")
+    .query("sen_length < 300")
+)
+total_candidates_df.head(2)
```
ajlee21, line 86: Is there a reason why the "disease" column is NaN.  This table is populated based on your tagger?

compound_disease/compound_treats_disease/word_vector_experiment/generate_word_vectors.py
```diff
@@ -0,0 +1,244 @@
...
+    .reset_index()
+    .rename({"index":"word", 0:"index"}, axis=1)
+)
+word_dict_df.to_csv("results/compound_treats_disease_word_dict.tsv", sep="\t", index=False)
+word_dict_df.head(2)
+
+
+# # Embed all of Compound Treats Disease Sentences
+
+# **Note**: Must run this section separately because the kernel cannot handle both training the word vectors and then embedding each CtD sentence.
```
ajlee21: If this is the case, does it make more sense to have separate notebooks for the training and embedding?

compound_disease/compound_treats_disease/word_vector_experiment/generate_word_vectors.py
```diff
@@ -0,0 +1,244 @@
...
+# 
+# This section embesd all candidate sentences. For each sentence, we place tags around each mention, tokenized the sentence and then matched each token to their corresponding word index. Any words missing from our vocab receive a index of 1. Lastly, the embedded sentences are exported as a sparse dataframe.
```
ajlee21: typo: embeds

compound_disease/compound_treats_disease/word_vector_experiment/generate_word_vectors.py
```diff
@@ -0,0 +1,244 @@
...
+    pd
+    .read_table("../dataset_statistics/results/all_ctd_map.tsv.xz")
+    .query("sen_length < 300")
+)
+total_candidates_df.head(2)
+
+
+# # Train Word Vectors
+
+# This section trains the word vectors using the specifications described above.
```
ajlee21, line 91: Not sure how necessary this is because perhaps future people looking at this PR will be more familiar with NLP, but I am a bit lost about what is the input and output of your training and embedding steps.  I am also not sure how they fit into your bigger picture - how are the embeddings used downstream?

ajlee21 (approved):
Really nice PR.

Made a few comments on your compound-treats-disease nb that can be applied to the other relationship types.

compound_disease/compound_treats_disease/word_vector_experiment/generate_word_vectors.py
danich1: Good catch. Turns out the file I generated in a previous notebook added an extra column called disease. This column is replaced with doid_name. I wrote down to remove this column in a future PR.

compound_disease/compound_treats_disease/word_vector_experiment/generate_word_vectors.py
danich1: Will do. I like that idea more than what I was currently doing.

compound_disease/compound_treats_disease/word_vector_experiment/generate_word_vectors.py
danich1: **tl;dr** - for a computer to utilize sentences in a machine learning sentence we need to "embed" them into a vector space. Input: all individual words found in sentences -> output: vector space with dimensions [total vocab size, word vector size (typically somewhere between 50-300)]

The end goal is to train a model to classify sentences; however, a computer can only work with numbers and not words. To fix this problem one has to use a coding scheme to essentially "embed" sentences into a vector space a computer can use. 

One way is to do a one hot encoding where you have large dimensional vectors (the total vocab size that one is working with. i.e. 300k+) and place a 1 at the index that represents a given word. The downside as you can imagine is this method doesn't perform quite as well in some cases. 

Despite one hot embedding, another way to "embed" sentences is to use [word2vec](https://towardsdatascience.com/introduction-to-word-embedding-and-word2vec-652d0c2060fa) which is a fancy way of saying train a neural network to learn the context of sentences and then generate embeddings for each word.  <- The project works with this idea.

Ultimately, I use these word embeddings to train a convolutional neural network that will classify the sentences I am working with.

PR #85, opened 2019-07-19
Uploading final disc results

danich1:
This PR contains the final results for the discriminator model. Not much code review needed. Just take a look at the figures generated. Let me know what you think.

Side Note: Data Files for this pull request will show up in a new one. Don't want to overwhelm you with files.


ajlee21 (approved):
Very nice figures!

Just a few comments:
1.  For the dev set, looks like there is a very marginal difference in performance between the gen and disc model.  Not sure if this is expected.  I don't remember what the disc is adding to the sentence classification on top of the gen model.  This gap seems to increase using the test set though.

2.  For the dev set, it looks like the disc model performs slightly worse compared to the gen model for all predictions expect DaG.  But then in the test set its the CtD that seems to be the best prediction.  Not sure if you might know why this is?

danich1:
1. The discriminator model uses the output of the generative model to make sentence classifications. In other words-> the generative model gives each sentence a confidence score that is the likelihood of mentioning a relationship. Then the discriminator model adds sentences features on top of the generative model's output to theoretically improve the final confidence score. In a perfect world the discriminator should out perform the generator, but as you can see this process is quite messy.

2. That's a really great question. Not sure the concrete answer there, but my main hunch is that class imbalance plays an important role with performance. CbG and GiG only have about less than 10% of positive sentences labeled, while DaG has about 50-50 split. Another pitfall is that the evaluation sets could use more labeled examples. Ideally I should have about 1k sentences labeled for each relationship, but given the circumstances that is not a trivial thing to do. Lastly, some relationships are easier to predict than others. It could be that DaG and CtD are easier to detect than GiG and CbG.

PR #88, opened 2019-07-22
Literature model

danich1:
This PR implements and compares the CoCoScore text mining system against my current system. I used hetionet v1 edges for the comparison, where the positives are edges already established in hetionet and the negatives are edges not in hetionet v1. There might be a might be a confounding issue for precision-recall performance, since hetionet v1 is not a fully complete network (some non-zero edges might actually be true edges but are considered negative in this context).


compound_disease/compound_treats_disease/literature_models/coco_score/coco_score.py
```diff
@@ -0,0 +1,376 @@
+
+# coding: utf-8
+
+# # CoCoScore Implementation
+
+# This notebook consists of implementing the [CoCoScore](https://www.biorxiv.org/content/10.1101/444398v1) literature model for comparison.
```
ajlee21: Might have a sentence that describes the metric in addition to the hyperlink

compound_disease/compound_treats_disease/literature_models/coco_score/coco_score.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+all_embedded_cd_df = pd.read_table("../../word_vector_experiment/results/all_embedded_cd_sentences.tsv.xz")
+all_embedded_cd_df.head(2)
+
+
+# In[11]:
+
+
+word_vectors = pd.read_csv(
```
ajlee21, line 126: Why is the first row blank?

compound_disease/compound_treats_disease/literature_models/coco_score/coco_score.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+# In[25]:
+
+
+precision, recall, _ = precision_recall_curve(
+    total_dev_sentences_df.curated_ctd,
+    pd.read_table('dev_predictions.tsv', header=None, sep=" ")[1]
+)
+auc(recall, precision)
```
ajlee21, line 248: So these curves are how accurately you were able to predict an edge in hetionet?

ajlee21 (approved):
Looks good.  Just a few comments that you can feel free to get to after your manuscript is due, as discussed.

So for the .png file, you're assuming CoCo score is the baseline that you want to compare you're relationships to?

compound_disease/compound_treats_disease/literature_models/coco_score/coco_score.py
danich1: Good catch that is a typo that I'll make sure to fix.

compound_disease/compound_treats_disease/literature_models/coco_score/coco_score.py
danich1: Exactly. These curves show the varying recall and precision thresholds for each cutoff score. The better the model the higher the area under the curve is.

danich1:
> So for the .png file, you're assuming CoCoScore is the baseline that you want to compare you're relationships to?

The .png file compares our model with the CoCoScore literature model. For this particular instance CoCoScore is our "baseline" model. The goal here is to show that our method performs just as good as a literature model.

PR #89, opened 2019-07-26
Error Analysis and Few Other Results

danich1:
This pr contains three modules: 
1. a random label function that emits positive and negative labels at a specified frequency. 
2. Edge predictions reporting the test set only 
3. Model calibration tables for each edge type.

No need to review code unless you want to.


ajlee21 (approved):
Figures look good.  I know you're in a rush to get alot done for your manuscript to let me know if you'd like me to take a look at more code later.

The one comment I had is that I'm a bit confused about the y-axis label for your polarity figures.  So is the interpretation of the curve that for a given number of DaG label functions I have an average confidence of 45% that these sentences are a positive association?  Or is that the percentage of sentences that are pos of the total?

danich1:
Ah. For that figure I am incorporating a single label function onto the baseline model. This single label function randomly generates labels (pos or neg) at a certain frequency (x-axis). For example this mentioned function could emit negative labels for 25% of the training set or for 50% of the training set. I did this experiment to understand why the Generative model performance increases for one label function and then drastically decreases.

PR #90, opened 2019-08-06
Quick fix

danich1:
This PR updates the plots for my manuscript.  Should be a quick review.


ajlee21 (approved):
Really nice figures, just a few questions for my own understanding and curiosity:

1.  For the "Reconstructing Hetionet" figure, what is considered a TP here?  So does this mean that we have very little confidence that any GiG relationships are true that exist in hetionet?

2.  This may have already been discussed amongst you and Casey and Daniel, but I wonder if the "0-Only Uses Relational.." could be put in the caption and doesn't need to be in the figure?

3.  Again, this might have already be discussed, in which case you can ignore this comment.  But I wonder if there is an alternative name for "CoCoScore" because my understanding is that both the "Our model" and "CoCoScore" are measuring the co-occurrence score but with different models, so calling one "CoCoScore" seems confusing.  

The figures looks very professional!  Great job!

danich1:
1. Edges already existing in Hetionet are considered the true positives. No that is too strong of a statement. It is better to say that our model has more uncertainty when predicting gene interacts gene edges.

2. For the sake of PSB manuscript leaving the caption in is the best option. Ultimately, I can go either way on this.

3. CoCoScore is actually the published baseline model's name. Cant remove or change it. "Our Model" was suppose to be changed, but I missed making that change before submission. Also, these methods are doing **very different** things to classify edges. 

Our Model -> predict edges by classifying sentences aka supervised machine learning. 
CoCoScore -> a hybrid approach using co-occurrence scores and logistic regression.

PR #92, opened 2019-08-16
Fixes lingering outdated urls

danich1:
This PR fixes some of the outdated urls used in my notebooks. I also have moved and removed files for clarity. Need to make one more pass through the folders before review.


danich1:
Feel free to take a look when u get the chance.

ajlee21 (approved):
Nice clean nbs

For the data_generator nbs I think that adding some comments about what the different tables are would make it faster for future you/others to read.  

FOr tables that have NaN values, are those because that data is not available in the DB or because you weren't able to import it?

The statistics nbs are very well documented!

danich1:
The tables that have a NaN value are candidates that aren't contained in Hetionet. The column that has NaN should be the source column that represents the source of an edge.

PR #93, opened 2019-11-05
Updating Snorkel and Environment

danich1:
This PR starts off post PSB rejection and the overall goal here is to update the snorkel package and other outdated packages.

Reasons for updating:
- More stable Generative Model. The parameter the model estimates is mu that represents the likelihood of a label function emitting a label (abstain, negative, positive) given the true class.
Before Upgrade: 
 mu - [
    [ 0.106, -0.0014],
    [0.000, 0.000],
    [0.000, -0.001]
 ]
After Upgrade:
  mu - [
        [1.0000e-07, 1.0000e-07],
        [4.1568e-03, 4.5846e-03],
        [1.0000e-07, 1.0000e-07]
 ]
^ floating point calculations better reflect probability estimates.
-  Able to obtain weights for each label function. This provides better insight to what label functions the model considers important
- Code style has typing which makes reading/using their code easier
- Overdue task since I have to retrace my steps on this project.

Adding @dhimmel to keep him up to date while he is afk during the next few days. (Another PR will come shortly)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,46 +1,47 @@
-name: snorkeling
+name: snorkeling_advanced
 channels:
   - conda-forge
   - pytorch
 dependencies:
 - backports.lzma==0.0.8
 - beautifulsoup4=4.6.0
-- gensim=3.6.0
-- ipykernel=4.8.2
-- ipywidgets=7.1.0
+- gensim=3.8.1
+- ipykernel=5.1.2
+- ipywidgets=7.5.1
 - jupyter=1.0.0
-- jupyter_client=5.0.1
-- jupyter_console=5.2.0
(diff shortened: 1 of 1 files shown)


environment.yml
```diff
@@ -5,42 +5,43 @@ channels:
 dependencies:
 - backports.lzma==0.0.8
```
dhimmel: this can probably go now that you're using python 3

dhimmel (approved):
If you can install the environment and it works... then it's good.

ajlee21 (commented):
Looks good to me.  I think I will defer to @dhimmel for his knowledge on conda environments.  One question is why you think the values for mu look "better" compared to earlier?  Is smaller better?

PR #95, opened 2019-11-11
Updated gen datafiles

danich1:
This PR follows #94. All this PR does is remove old data files and adds new data files. Nothing to review here just approve after reviewing the previous PR.

PR #94, opened 2019-11-11
Updated gen model

danich1:
This PR follows the previous pr (#93). I took the time to carefully examine my code and notice a few bugs concerning the generative model performance:
1. The grid search algorithm wasn't selecting the best model. It was selecting a model at random. Fixed that and then I realized I was using the wrong column to estimate model performance. 
2. The indices for the GiG and possibly CbG relation had an added label function that shouldn't be in there.

After fixing the above bugs I have corrected the gen model's auroc and aupr point plot grid:

| Before Bug Fix | After Bug Fix |
| --- | --- |
| [tune set AUROC](https://raw.githubusercontent.com/greenelab/snorkeling/dd70fd47933e3ceebbef1c7f4990454ff4807904/figures/label_sampling_experiment/transfer_dev_set_auroc.png) | [tune set AUROC](https://raw.githubusercontent.com/danich1/snorkeling/52a2ae71fbc82159342bc68b1712a38ecf53ff45/figures/label_sampling_experiment/transfer_dev_set_auroc.png)  | 
| [tune set AUPR](https://raw.githubusercontent.com/greenelab/snorkeling/dd70fd47933e3ceebbef1c7f4990454ff4807904/figures/label_sampling_experiment/transfer_dev_set_auprc.png)  | [tune set AUPR](https://raw.githubusercontent.com/danich1/snorkeling/52a2ae71fbc82159342bc68b1712a38ecf53ff45/figures/label_sampling_experiment/transfer_dev_set_auprc.png) |
| [test set AUROC](https://raw.githubusercontent.com/greenelab/snorkeling/dd70fd47933e3ceebbef1c7f4990454ff4807904/figures/label_sampling_experiment/transfer_test_set_auroc.png) | [test set AUROC](https://raw.githubusercontent.com/danich1/snorkeling/52a2ae71fbc82159342bc68b1712a38ecf53ff45/figures/label_sampling_experiment/transfer_test_set_auroc.png) | 
| [tune set AUPR](https://raw.githubusercontent.com/greenelab/snorkeling/dd70fd47933e3ceebbef1c7f4990454ff4807904/figures/label_sampling_experiment/transfer_test_set_auprc.png) | [test set AUPR](https://raw.githubusercontent.com/danich1/snorkeling/52a2ae71fbc82159342bc68b1712a38ecf53ff45/figures/label_sampling_experiment/transfer_test_set_auprc.png) | 

[...]

PR #96, opened 2019-12-12
Word vector fix

danich1:
This PR follows #94 and #95. After the generative model was fixed, the next step is to fix the issue contained in the word vectors and sentence embeddings. Long story short I had an index error when trying to load sentences for the discriminator model.

e.g. 
Sentence: `Over-expression in disease`  
0 - represent the word `over-expression`,  1 - represents the word `disease` and 2- represents the word `in`

This is problematic because 0 and 1 are reserved for `null` character and `unknown token` character. This means two important words were getting drowned out for null's and unknowns, which lead to poor discriminative model performance. 

Issue is fixed and now the disc model is working. Results will appear in next PR.


gene_gene/gene_interacts_gene/word_vector_experiment/embed_sentences.py
```diff
@@ -85,11 +85,12 @@
...
 # This section embeds all candidate sentences. For each sentence, we place tags around each mention, tokenized the sentence and then matched each token to their corresponding word index. Any words missing from our vocab receive a index of 1. Lastly, the embedded sentences are exported as a sparse dataframe.
 
-# In[6]:
+# In[ ]:
 
 
-word_dict_df = pd.read_table("results/gene_interacts_gene_word_dict.tsv")
+word_dict_df = pd.read_table("output/gene_interacts_gene_word_dict.tsv")
 word_dict = {word[0]:word[1] for word in word_dict_df.values.tolist()}
+fixed_word_dict = {word:word_dict[word] + 2 for word in word_dict}
```
dhimmel, line 93: so this is the big fix?

dhimmel (approved):

gene_gene/gene_interacts_gene/word_vector_experiment/embed_sentences.py
danich1: Yeah that is the bug fix.

PR #99, opened 2020-01-06
Model calibration

danich1:
This PR shows the results of calibrating the discriminator models. Deep Learning models tend to be over confident in their predictions. As a consequence there could be an inflation of high scores for sentences and vise versa. To resolve this I used a [temperature scaling approach](https://github.com/gpleiss/temperature_scaling). Each notebook contains a figure showing the results after calibration.

For swiftness of reviewing this PR just look at the notebooks.


dhimmel (approved):

ajlee21 (commented):
This is interesting.

Just a couple of questions out of curiosity:
1.  So your "actual score" is your manually curated score for a sentence?  So its a boolean?
2.  Based on the line graph in your nb it looks like even after calibration that your "predicted score" is still always higher than your "actual score".  Is the calibration meant to get it so that actual = predicted?

danich1:
1. Close. In a calibration curve the actual score is the fraction of positive labels located in that bin For example at 0.8 approximately 80% of the bin belongs to the positive class.This [resource](https://scikit-learn.org/stable/modules/calibration.html) provides a bit more intuition on how calibration works.

2. Right. The idea is to get the dotted line to line up with the dashed line as close as possible. By accomplishing then we can argue the model's predictions are more reliable.

PR #102, opened 2020-03-10
Finalize database insertion

danich1:
This PR adds code to perform database insertion and updates a few old notebooks to reflect current results.
Main file to look at is pubtator-to-postgres notebook.


create_database/pubtator-to-postgres.py
```diff
@@ -0,0 +1,392 @@
...
+
+from sqlalchemy import func
+from string import punctuation
+import lxml.etree as et
+
+
+# In[ ]:
+
+
+def offsets_to_token(left, right, offset_array, lemmas, punc=set(punctuation)):
```
ajlee21: Would it make more sense to have these scripts be outside of the nb?

create_database/pubtator-to-postgres.py
```diff
@@ -0,0 +1,392 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# # Inserting Pubtator into a Database
+
+# This notebook is designed to load each [Pubmed](https://www.ncbi.nlm.nih.gov/pubmed/) abstract. It uses our own [pubtator repository](https://github.com/greenelab/pubtator) to convert each abstract into the appropriate format (xml) to load into a postgres database. Run the pubtator scripts before running this notebook, so the pubtator xml file can be constructed. Once constructed this notebook is designed to parse the data into a database. After loading each abstract, candidate extraction is performed for each relationship type.
+
+# ## MUST RUN AT THE START OF EVERYTHING
```
ajlee21, line 8: I am having this same issue currently so I'm not sure the best solution for this.  But since you only have to run this top set of scripts once.  Does it make more sense to have it in its own separate nb?  Or have a wrapper script to call these set of scripts and keep track of if was already run based on the output?

create_database/pubtator-to-postgres.py
```diff
@@ -0,0 +1,392 @@
...
+# In[ ]:
+
+
+category_list = np.random.choice([0,1,2], total_sentences, p=[0.7,0.2,0.1])
+
+
+# In[ ]:
+
+
+# Divide the sentences into train, dev and test sets
```
ajlee21, line 163: So you are not grouping sentences based on document?  (i.e. all the sentences from document A are not in training but can be split across training and testing?)  I wonder if this will make a difference?  I feel like you might lose some context if they are not grouped, but I also understand wanting to make sure each group is similar in what it has seen

ajlee21 (approved):
Very well documented code!  So much great progress!

Just some structural comments that you can decide if/when is an appropriate time to address them.

create_database/pubtator-to-postgres.py
danich1: Yes I agree. I'll move the functions into a separate script.

create_database/pubtator-to-postgres.py
danich1: Well it depends on the design. If the imports for each notebook are the same, then it makes sense to move the code into a separate script and then just have the notebooks call that import script. In this situation this is the only time I need to have the following code as future work won't even use this notebook. I just have it here for project completion sake.

create_database/pubtator-to-postgres.py
danich1: Right each sentence is considered its own individual entity. Also, you are correct. The biggest issue with this project was that I lost contextual information, which made sentence prediction extremely difficult (e.g. ambiguous sentences). Something to keep in mind when I move towards full text.

PR #103, opened 2020-03-10
Finalize figure generator

danich1:
This PR updates the figure generation code for the snorkeling manuscript. Intense code review is not needed as you have seen this code before.


ajlee21 (approved):
Looks good.  Just curious, it looks like you updated the entire file.  Is there a reason for that?

danich1:
Discussed with @ajlee21 and realized the file she is referring to is the `disc_model_figure_generator.py`. Instead of having one large notebook that plots the generator model's performance and discriminator model's performance, I decided to separate the notebook. As a consequence this file: `disc_model_figure_generator.py` looks new.

PR #104, opened 2020-03-10
Finalize expanding hetionet

danich1:
This PR contains updated literature model performance and my model performance.


compound_disease/compound_treats_disease/edge_prediction_experiment/edge_prediction_experiment.py
```diff
@@ -82,7 +117,7 @@
 grouped_candidates_pred_df.head(2)
 
 
-# In[6]:
+# In[9]:
```
ajlee21, line 121: Just to clarify.  For a given gene A and compound B pair, you may have multiple sentences that co-mention these.  And each sentence will have a ```pred``` score that indicates the confidence of compound B binding gene A.  These scores can be aggregated via the mean, median or max?

compound_disease/compound_treats_disease/edge_prediction_experiment/edge_prediction_experiment.py
```diff
@@ -115,13 +150,13 @@
 
 # This section aims to answer the question: What metric (Mean, Max, Median) best predicts Hetionet Edges?
 
-# In[9]:
+# In[12]:
```
ajlee21, line 154: The gold standard is based on your hand-labeled set?

compound_disease/compound_treats_disease/edge_prediction_experiment/edge_prediction_experiment.py
```diff
@@ -338,7 +453,7 @@
 edges_df
 
 
-# In[20]:
+# In[26]:
```
ajlee21, line 431: Minor -- move import to the top of the nb

compound_disease/compound_treats_disease/edge_prediction_experiment/edge_prediction_experiment.py
```diff
@@ -283,31 +318,111 @@
...
 # # Total Recalled Edges
 
-# How many edges of hetionet can we recall using a cutoff score of 0.5?
+# How many edges of hetionet can we recall using an equal error rate cutoff score?
 
-# In[19]:
+# In[22]:
+
+
+def tag_sentence(x):
```
ajlee21: Maybe have this be a standalone script since several notebooks are calling it too

compound_disease/compound_treats_disease/edge_prediction_experiment/edge_prediction_experiment.py
```diff
@@ -283,31 +318,111 @@
...
+# In[21]:
 
 
 edges_added_df.to_csv("output/precision_ctd_edges_added.tsv", index=False, sep="\t")
 
 
 # # Total Recalled Edges
 
-# How many edges of hetionet can we recall using a cutoff score of 0.5?
+# How many edges of hetionet can we recall using an equal error rate cutoff score?
```
ajlee21, line 337: Not exactly clear on what the below scripts are doing.  What do you mean by "equal error rate cutoff"?  It sounds like you are trying to see how many edges are in hetionet, but that was already shown above.  I think also the word "recall" is making me think that you're specifically looking at recall scores of sentences

ajlee21 (approved):
Looks pretty good to me.  Just some points of clarity.

This might be a moot point because you're so far along on this project.  But I think you might have the same situation as me where you have duplicated scripts for each relationship type and could try to have a set of shared scripts that each relationship calls.  I'm not sure if its something where you have thought about already so maybe it won't make sense in your case.

compound_disease/compound_treats_disease/edge_prediction_experiment/edge_prediction_experiment.py
danich1: Bingo you are correct.

compound_disease/compound_treats_disease/edge_prediction_experiment/edge_prediction_experiment.py
danich1: In this case no the gold standard are the edges Hetionet already contains. When predicting sentences the gold standard is my hand-labeled set.

compound_disease/compound_treats_disease/edge_prediction_experiment/edge_prediction_experiment.py
danich1: This section takes a different view at finding the optimal cutoff. Instead of use precision-recall curves I used AUROC curves instead. In this case equal error rate means I want a ROC cutoff point that gives me the same number of false positives as false negatives (equal error). I was exploring another route to call Hetionet edges.

